Issue #634, opened 2014-02-02
fork existing repository


tillmo:
There should be the possibility to fork an existing git or svn repository without creating a mirror.

Issue #351, opened 2013-11-01
Allow switch from mirroring repositories to normal


0robustus1:
We should allow (at least admins) to switch a repository from *mirroring* mode to
the normal repository mode (with push support and so on).

PR #1030, opened 2014-09-04
634 fork and unmirror repository

eugenk:
Shall fix both issues of #634 (and #351): Allowing to fork a repository and converting a mirror to a local repository.

This also improves the javascript in the new repository view as it now updates everytime the source address is changed (key press, click, cut, paste) and not only when its input field loses focus.


ebolloff:
:+1:

PR #1053, opened 2014-09-22
Basic provisioning working with Chef and Vagrant; much left to do

jamesaoverton:
Here are some scripts and documentation for using Chef (and optionally Vagrant) to install Ontohub. See issue #785.

The system is not yet complete. It installs all the dependencies and runs "bundle install" but does not configure the database or do many of the other things in the deployment guide. Still, I hope it's helpful.

### DIFF
--- a/provisioning/Cheffile
+++ b/provisioning/Cheffile
@@ -1,0 +1,9 @@
+site "http://community.opscode.com/api/v1"
+
+cookbook 'apt'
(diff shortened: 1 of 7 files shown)


eugenk:
Now that we have the new server and someone who takes care of the ontohub deployment to it, we don't need this as a starting point to setting it up with chef.

Issue #958, opened 2014-07-18
Handle git rm


eugenk:
You can delete files containing ontologies via git, but ontohub doesn't handle this case (even though it is detected).

What shall happen with the ontology object (and maybe importing ones) in the database? Shall the importing ontologies be re-parsed?

eugenk:
Closed by #1075

PR #1075, opened 2014-10-08
617 handle file deletion

### LINKED ISSUE #617: deleting files/repositories
- deleting a file in a repository seems not possible via the GUI. 

- after going to the ontology file, a delete button appears top right corner (probably admin only) - this deletes the entire repository without warning, ie it does not say what will be deleted.

eugenk:
This shall fix #617 and #958.

Files in the repository can be deleted now via the web interface. 
Then, their ontologies will be marked as 'not having a file' (`has_file`).
This flag is stored in the database, but only for the head-version of an ontology (as a cache).
One can still ask for a specific commit, if an ontology has a file (`has_file(commit_oid)`), which then accesses the hard drive.

Deletion via ssh is detected and the flag is set in that case.

In the view (ontology#show), the nonexistence of a corresponding file is indicated by text and the absence of a link to the file.

### DIFF
--- a/app/controllers/files_controller.rb
+++ b/app/controllers/files_controller.rb
@@ -49,11 +49,18 @@ class FilesController < InheritedResources::Base
   def update
     @repository_file = resource_class.create(params.merge(user: current_user))
     if resource.valid?
-      flash[:success] = "Successfully changed the file."
+      flash[:success] = "Successfully changed the file #{params[:path]}."
       redirect_to fancy_repository_path(repository, path: resource.target_path)
     else
       render :show
     end
+  end
+
+  def destroy
+    repository.delete_file(params[:path], current_user)
+    flash[:success] = "Successfully deleted the file #{params[:path]}."
+    redirect_to(fancy_repository_path(repository,
+      path: repository.deepest_existing_dir(params[:path])))
   end
 
   protected
--- a/app/helpers/ontology_helper.rb
+++ b/app/helpers/ontology_helper.rb
@@ -56,7 +56,8 @@ module OntologyHelper
   end
 
   def last_file_path(resource)
-    repository_ref_path(repository_id: resource.repository.to_param,
+    repository_ref_path(
+      repository_id: resource.repository.to_param,
       ref: resource.current_version.commit_oid,
       path: resource.current_version.path,
       action: :show)
--- a/app/helpers/ontology_helper.rb
+++ b/app/helpers/ontology_helper.rb
@@ -13,4 +13,52 @@ module OntologyHelper
   def show_evaluate?
     show_oops? #|| show_foo?
   end
+
+  def status_tag(resource)
+    version = resource.is_a?(Ontology) ? resource.current_version : resource
+    uri = repository_ontology_ontology_version_url(
+      version.ontology.repository, version.ontology, version)
+    html_opts = {
+      class: 'ontology-version-state',
+      data: {
+        ontology_version_id: version.id,
+        uri: uri,
+        state: version.state,
+      }
+    }
+    content_tag(:small, html_opts) do
+      status(version)
+    end
+  end
+
+  def status(resource)
+    html = content_tag :span, resource.state
+
+    if %w(pending fetching processing).include? resource.state
+      html << " " << image_tag('spinner-16x16.gif', class: 'spinner')
+    end
+
+    if resource.state == 'failed' and resource.is_a? Ontology
+      version = resource.versions.last
+
+      link = ' ('
+      link << link_to('error',
+        [resource.repository, resource, :ontology_versions],
+        :'data-original-title' => version.last_error,
+        class: 'help'
+      )
+      link << ')'
+
+      html << content_tag(:span, link.html_safe, class: 'error')
+    end
+
+    html
+  end
+
+  def last_file_path(resource)
+    repository_ref_path(repository_id: resource.repository.to_param,
+      ref: resource.current_version.commit_oid,
+      path: resource.current_version.path,
+      action: :show)
+  end
 end
--- a/app/helpers/ontology_helper.rb
+++ b/app/helpers/ontology_helper.rb
@@ -14,47 +14,6 @@ module OntologyHelper
     show_oops? #|| show_foo?
   end
 
-  def status_tag(resource)
-    version = resource.is_a?(Ontology) ? resource.current_version : resource
-    uri = repository_ontology_ontology_version_url(
-      version.ontology.repository, version.ontology, version)
-    html_opts = {
-      class: 'ontology-version-state',
-      data: {
-        ontology_version_id: version.id,
-        uri: uri,
-        state: version.state,
-      }
-    }
-    content_tag(:small, html_opts) do
-      status(version)
-    end
-  end
-
-  def status(resource)
-    html = content_tag :span, resource.state
-
-    if %w(pending fetching processing).include? resource.state
-      html << " " << image_tag('spinner-16x16.gif', class: 'spinner')
-    end
-
-    if resource.state == 'failed' and resource.is_a? Ontology
-      version = resource.versions.last
-
-      link = ' ('
-      link << link_to('error',
-        [resource.repository, resource, :ontology_versions],
-        :'data-original-title' => version.last_error,
-        class: 'help'
-      )
-      link << ')'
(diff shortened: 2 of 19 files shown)


app/controllers/files_controller.rb
```diff
@@ -49,13 +49,20 @@ def create
...
     if resource.valid?
-      flash[:success] = "Successfully changed the file."
+      flash[:success] = "Successfully changed the file #{params[:path]}."
       redirect_to fancy_repository_path(repository, path: resource.target_path)
     else
       render :show
     end
   end
 
+  def destroy
```
eugenk: I'm not sure if this is a hound/rubocop bug or I just don't understand this comment. There is no branching in this method.

app/controllers/files_controller.rb
```diff
@@ -49,13 +49,20 @@ def create
...
+    repository.delete_file(params[:path], current_user)
+    flash[:success] = "Successfully deleted the file #{params[:path]}."
+    redirect_to(fancy_repository_path(repository,
+      path: repository.deepest_existing_dir(params[:path])))
```
eugenk: I think this style is valid in this case because the indentation would be way to much. If the reviewer objects, I'm willing to change it though.

app/models/ontology.rb
```diff
@@ -274,6 +275,23 @@ def formality_levels
...
+  def file_deleted?(commit_oid = nil)
+    !has_file?(commit_oid)
+  end
+
+  # alias_method doesn't work for this one.
+  def has_file?(commit_oid = nil)
+    has_file(commit_oid)
+  end
+
+  def has_file(commit_oid = nil)
```
eugenk: No thanks. The method right above this one takes care of that. Here, I overwrite the attribute getter.

eugenk:
I'm ignoring the hound comments on `routes.rb`/spacing because then the file would look more inconsistent. The double quotes comments are just wrong. Seems to be a bug in hound/rubocop.

0robustus1:
please rebase staging.

eugenk:
Rebased.

app/models/repository/git.rb
```diff
@@ -222,6 +225,15 @@ def priority_settings
     @priority_settings ||= OpenStruct.new(Settings.git[:push_priority])
   end
 
+  def mark_ontology_as_having_file(path, has_file)
```
0robustus1: Maybe use the keyword-syntax for `has_file` as this shows better what the boolean stands for in calls. e.g.
`mark_ontology_as_having_file(path, has_file: true)` is more readable than `mark_ontology_as_having_file(path, true)`.

eugenk: Done.

app/helpers/ontology_helper.rb
```diff
@@ -13,4 +13,53 @@ def download_path(resource)
   def show_evaluate?
     show_oops? #|| show_foo?
   end
+
+  def status_tag(resource)
```
0robustus1: This method seems to be unused (a `grep` couldn't find any other occurence other than the definition itself).

app/helpers/ontology_helper.rb
```diff
@@ -13,4 +13,53 @@ def download_path(resource)
...
+        uri: uri,
+        state: version.state,
+      }
+    }
+    content_tag(:small, html_opts) do
+      status(version)
+    end
+  end
+
+  def status(resource)
```
0robustus1: The same as above seems to apply here. The only usage i could find was the one inside the `status_tag` method, which also seems to be unused.

app/models/ontology.rb
```diff
@@ -195,6 +195,23 @@ def versioned_locid
     current_version.locid
   end
 
+  def file_deleted?(commit_oid = nil)
+    !has_file?(commit_oid)
+  end
+
+  # alias_method doesn't work for this one.
```
0robustus1: just for my information, why doesn't `alias_method` work?

app/models/ontology.rb
```diff
@@ -195,6 +195,23 @@ def versioned_locid
     current_version.locid
   end
 
+  def file_deleted?(commit_oid = nil)
```
0robustus1: You could document the meaning of `commit_oid` for these methods. I assume this checks if the file doesn't exist in the state of this commit (and defaults to HEAD if `commit_oid` is `nil`)?

app/models/repository/git.rb
```diff
@@ -4,8 +4,8 @@ module Repository::Git
   extend ActiveSupport::Concern
 
   delegate :commit_id, :dir?, :empty?, :get_file, :has_changed?,
-    :is_head?, :path_exists?, :paths_starting_with, :points_through_file?,
-    to: :git
+           :is_head?, :path_exists?, :paths_starting_with,
+           :points_through_file?, :deepest_existing_dir, to: :git
```
0robustus1: maybe a new line for the `to: :git` as it is the important part?

lib/git_repository/history.rb
```diff
@@ -70,6 +72,17 @@ def commits(start_oid: nil, stop_oid: nil, path: nil, limit: nil, offset: 0, wal
     end
   end
 
+  def rugged_walk_order(walk_order)
```
0robustus1: I personally prefer to use constant-hashes instead of case-statements if possible. What do you think?

lib/git_repository/history.rb
```diff
@@ -59,7 +59,9 @@ def commits(start_oid: nil, stop_oid: nil, path: nil, limit: nil, offset: 0, wal
     stop_oid = nil if stop_oid =~ /\A0+\z/
 
     walker = Rugged::Walker.new(@repo)
-    walker.sorting(walk_order) if walk_order
+    if rwo = rugged_walk_order(walk_order)
```
0robustus1: Wow the one case where you actually want the `=` in the `if`-condition ;)

spec/models/repository/save_and_delete_file_spec.rb
```diff
@@ -50,33 +53,57 @@
...
+        tmpfile
+      end
+
+      before do
+        repository.save_file(file_path, target_path, message, user)
+        repository.save_file(file_path2, target_path, message, user)
+      end
+
+      it 'create a new ontology version' do
+        expect(repository.ontologies.
```
0robustus1: Indentation and line-breaks seem strange.

spec/models/repository/save_and_delete_file_spec.rb
```diff
@@ -50,33 +53,57 @@
...
-          repository.save_file(file_path, target_path, message, user)
-          repository.save_file(file_path2, target_path, message, user)
-        end
-
-        it 'create a noew ontology version' do
-          expect(repository.ontologies.
-            where(basepath: File.basepath(target_path)).first!.versions.count).
-            to eq(2)
-        end
+          to(change { Worker.jobs.count }.by(1))
```
0robustus1: as far as i know, you do not need the parentheses around the `to` arg.

app/helpers/ontology_helper.rb
eugenk: This is quite old code that has been changed (also the method has been renamed to `state_tag`) in several other pull requests. Seems like it got back in with a rebase.

app/models/ontology.rb
eugenk: If we use `alias_method` instead, we get this error:

> The single-table inheritance mechanism failed to locate the subclass: 'SingleOntology'. This error is raised because the column 'type' is reserved for storing the class in case of inheritance. Please rename this column if you didn't intend it to be used for storing the inheritance class or overwrite Ontology.inheritance_column to use another column for that information.

So, I assume `alias_method` is bad for the combination STI + aliasing an `attr_accessible`.

spec/models/repository/save_and_delete_file_spec.rb
eugenk: Fixed with 69b618d.

eugenk:
Tests are red because #1343's commits are not in here.

0robustus1:
So tests should be green if i merge #1343 locally and then run the tests?

eugenk:
Yes. And if you have the latest hets version (v0.99, 1430921534), of course.

0robustus1:
:+1:

PR #1111, opened 2014-11-05
1087 correct indefinite article for oms

hardik101:
added new gem and used its method 'with_indefinite_article' for all the found instances where it can be applied.


eugenk:
I just rebased staging. In my opinion, it's a :+1:

PR #1126, opened 2014-11-29
1122 remove dead code on bulk uploads

### LINKED ISSUE #1122: Remove dead code on bulk uploads
There still is code left on bulk uploads. The controller actions and routes are removed, but there are still some views and maybe other code left.

`OntologyVersion.source_url` is the attribute that was used in the bulk upload. On every machine ({,staging.,develop.}ontohub.org) we get:
```ruby
OntologyVersion.where('source_url IS NOT NULL')
=> []
```
So this one seems to be useless as well.

eugenk:
Shall fix #1122.


ebolloff:
:+1:

Issue #1066, opened 2014-10-05
provide rdf/xml representation of logics


tillmo:
When a logic is downloaded with mime type application/rdf+xml, Ontohub should repsond with an xml/rdf file representing the logic, see items with tag dol:logic at https://github.com/ontohub/OOR_Ontohub_API/blob/master/lola/registry/registry.rdf
The corresponding ontology can be found at https://github.com/ontohub/OOR_Ontohub_API/blob/master/lola/ontology/dol-rdf.omn

tillmo:
I would like to see a sample XML output. How is this possible?
```
wget --header 'Accept: application/rdf+xml' -O - http://develop.ontohub.org/logics/OWL
```
does not work...

0robustus1:
Both `curl -XGET -H "Accept: application/xml" http://develop.ontohub.org/logics/OWL` and `curl -XGET http://develop.ontohub.org/logics/OWL.xml` should work. I can add 'application/rdf+xml' as possible Accept-Header. However all three version would return a Content-Type of 'application/rdf+xml'.

tillmo:
@clange, could you please have a look at the output of
```
wget --header 'Accept: application/rdf+xml' -O - http://develop.ontohub.org/logics/OWL
```

clange:
Sorry that it took me a while. First two minor syntactic concerns. Use the RDF/XML validator at http://www.w3.org/RDF/Validator/ to see what's wrong. Or don't output RDF/XML yourself, but output the more human-friendly Turtle and use some standard library to convert it to RDF/XML, or employ a high-level library to generate, in the Ontohub code, "RDF triples", and leave it to the library to output these triples in any concrete syntax. RDF/XML is the most RDF serialization that's most widely supported by software, but at the same time the least convenient for humans. The unnecessary complexity of RDF/XML causes a lot of cognitive overhead particularly if you are not fully familiar with the RDF data model.

So, the declaration of the `xmlns:xsd` prefix is broken, and the following syntax is also wrong:

    <stand:standardizationStatus rdf:resource='http://purl.net/dol/1.0/standardization#Unofficial'>
    Unofficial
    </stand:standardizationStatus>

The object of a triple can either be a literal (string, number, etc.), which in RDF/XML is represented as the text content of an element. Or the object is a URI, which is what RDF/XML uses the `rdf:resource` attribute for.

[...]

Issue #32, opened 2012-03-29
Content negotiation between HTML, RDF/XML, plus optionally JSON and XML


clange:
First of all, which formats do we need, why do we need them, and what MIME types do they have?

* HTML (for the UI): text/html or application/xhtml+xml
* RDF/XML (we do not yet have this, but an RDF vocabulary for DOL is under development, and should be used by Ontohub): application/rdf+xml
* JSON (we get it for free from Rails, so let's serve it): application/json
* XML (we get it for free from Rails, so let's serve it): application/xml or text/xml.

Content negotiation works as follows: The client lists the requested MIME type in the Accept header of its HTTP request. If we have the format, we serve it. If we don't have it, we return "406 Unacceptable".

The serving should take place directly (unless linked data compliance requires a 303 redirect; see #33). Rails does support URLs with extensions such as ".json" (if I understood @corny correctly), but when some MIME type is requested from a URL without such an extension (e.g. /ontologies/MyOntology), we should serve it directly from this URL.

This applies to all relevant things (ontologies, entities, links, etc.) in Ontohub – whether their URIs are URLs within the Ontohub site (e.g. an ontology identified by the URI http://ontohub.informatik.uni-bremen.de/ontologies/Pizza), or whether we imported ontologies with external URIs into Ontohub.

Note that hardly any client will mention no MIME type explicitly. All modern browsers explicitly request HTML. Still, for clients such as wget, we should decide which format to serve by default.

PR #1127, opened 2014-11-30
1066 provide xml for logics

0robustus1:
Should at least be a partial-fix for #1066 and #32. It should be decided by @tillmo whether to merge this and if one or the other (or both) issues should be closed afterwards.


eugenk:
:+1: for the latest commits as well.

PR #1136, opened 2014-12-05
558 access token for private repositories

### LINKED ISSUE #558: Problem with private repositores - download of imported files leads to 403
When I use imports within a private repository, Hets fails to download the imported ontologies, because they are not accessible in a private repository:
```
wget http://ontohub.org/colore4/magma/semigroup.clif
--2014-01-11 19:51:07--  http://ontohub.org/colore4/magma/semigroup.clif
Auflösen des Hostnamen »ontohub.org (ontohub.org)«... 134.102.204.133
Verbindungsaufbau zu ontohub.org (ontohub.org)|134.102.204.133|:80... verbunden.
HTTP-Anforderung gesendet, warte auf Antwort... 403 Forbidden
2014-01-11 19:51:07 FEHLER 403: Forbidden.
```
Maybe a special password should be transfered in the URL?

eugenk:
This shall fix #558. As soon as hets (https://github.com/spechub/Hets/issues/1422) can handle it completely, this should work. Now, it is working for non-owl files whose file type can be detected by hets' magic.

After merging, don't forget to put this line into the crontab for the expired tokens to be deleted daily:
```
0 0  * * *  ~/current/script/rails runner 'AccessTokenDeletionWorker.new.perform'
```

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,6 +40,10 @@ class ApplicationController < ActionController::Base
 
   protected
 
+  def current_ability
+      @current_ability ||= Ability.new(current_user, params[:"access-token"])
+  end
+
   def authenticate_admin!
     unless admin?
       flash[:error] = 'you need admin privileges for this action'
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -41,7 +41,11 @@ class ApplicationController < ActionController::Base
   protected
 
   def current_ability
-      @current_ability ||= Ability.new(current_user, params[:"access-token"])
+    @current_ability ||= Ability.new(current_user, params[:'access-token'])
+  end
+
+  def params_to_pass_on_redirect
+    {:'access-token' => params[:'access-token']}
   end
 
   def authenticate_admin!
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,8 @@ class ApplicationController < ActionController::Base
   include PathHelpers
   include ApplicationHelper
 
+  PARAMS_TO_PASS_ON_REDIRECT = %i(access-token)
+
   # CanCan Authorization
   rescue_from CanCan::AccessDenied do |exception|
     if request.format.html?
@@ -45,9 +47,7 @@ class ApplicationController < ActionController::Base
   end
 
   def params_to_pass_on_redirect
-    new_params = {}
-    %i(access-token).each { |key| new_params[key] = params[key] if params[key] }
-    new_params
+    params.select { |k, _v| PARAMS_TO_PASS_ON_REDIRECT.include?(k) }
   end
 
   def authenticate_admin!
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -45,7 +45,9 @@ class ApplicationController < ActionController::Base
   end
 
   def params_to_pass_on_redirect
-    {:'access-token' => params[:'access-token']}
+    new_params = {}
+    %i(access-token).each { |key| new_params[key] = params[key] if params[key] }
+    new_params
   end
 
   def authenticate_admin!
--- a/app/models/ability.rb
+++ b/app/models/ability.rb
@@ -1,7 +1,7 @@
 class Ability
   include CanCan::Ability
 
-  def initialize(user)
+  def initialize(user, access_token)
     # Define abilities for the passed in user here.
 
     user ||= User.new # guest user (not logged in)
@@ -17,7 +17,10 @@ class Ability
         elsif subject.is_private
           subject.permission?(:reader, user) ||
           subject.permission?(:editor, user) ||
-          subject.permission?(:owner, user)
+          subject.permission?(:owner, user) ||
+          subject.access_tokens.unexpired.any? do |token|
+            token.to_s == access_token
+          end
         else
           true
         end
@@ -122,7 +125,10 @@ class Ability
       can :read, Category
     else
       can :show, Repository do |subject|
-        !subject.is_private && !subject.is_destroying
+        !subject.is_destroying && (!subject.is_private ||
+        subject.access_tokens.unexpired.any? do |token|
+          token.to_s == access_token
+        end)
       end
 
       can :read, Project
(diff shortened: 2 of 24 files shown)


eugenk:
We always pass an access token in the params. It has to be omitted when there is no access token

app/controllers/application_controller.rb
```diff
@@ -40,6 +40,14 @@ class ApplicationController < ActionController::Base
 
   protected
 
+  def current_ability
+    @current_ability ||= Ability.new(current_user, params[:'access-token'])
+  end
+
+  def params_to_pass_on_redirect
+    {:'access-token' => params[:'access-token']}
```
eugenk: No. This behavior is actually deactivated:
```yml
Style/SpaceInsideHashLiteralBraces:
  Description: "Use spaces inside hash literal braces - or don't."
  Enabled: true
  EnforcedStyle: no_space
```

app/models/ability.rb
```diff
@@ -15,7 +15,10 @@ def initialize(user)
         if subject.is_private
           subject.permission?(:reader, user) ||
           subject.permission?(:editor, user) ||
-          subject.permission?(:owner, user)
+          subject.permission?(:owner, user) ||
+          subject.access_tokens.unexpired.any? do |token|
```
eugenk: In this file: No thanks. This will be deleted soon, anyway.

eugenk:
in pause because it needs some rewriting after #1269 is merged (Hets::Options needs to be replaces)

eugenk:
This is now fully adjusted to the changes of interfacing with Hets. It is open for review again.

eugenk:
Oh and I rebased this branch on staging. In this process, the `Hets::Options` related commits were skipped (not needed any more, replaced by the new `HetsOptions`/`ProveOptions`).
Those skipped commits were the only ones with conflicts. The staging version has removed all the Hets-CLI related code and the `Hets::Options` related code from this branch changed those removed lines (and added - now useless - lines in between).

The first new commit is
Move class methods to the top.	f38207c

app/controllers/application_controller.rb
```diff
@@ -40,6 +40,16 @@ class ApplicationController < ActionController::Base
 
   protected
 
+  def current_ability
+    @current_ability ||= Ability.new(current_user, params[:'access-token'])
+  end
+
+  def params_to_pass_on_redirect
+    new_params = {}
+    %i(access-token).each { |key| new_params[key] = params[key] if params[key] }
```
0robustus1: How about using `select`?

eugenk: Right. `select` is better :D

app/models/repository/access.rb
```diff
@@ -4,6 +4,24 @@ module Repository::Access
   OPTIONS = %w[public_r public_rw private_r private_rw]
   DEFAULT_OPTION = OPTIONS[0]
 
+  def self.as_read_only(access)
+    access.split('_').first + '_r'
```
0robustus1: how about using a limit to split (e.g. `...split('_', 2)`) so optimizations can kick in? I know it's probably just a short string, but why not ;)

app/models/repository/access.rb
```diff
@@ -4,6 +4,24 @@ module Repository::Access
...
+  end
+
+  def destroy_expired_access_tokens
+    access_tokens.select(&:expired?).map(&:destroy)
```
0robustus1: Why `map`?

lib/hets/hets_options.rb
```diff
@@ -46,9 +47,19 @@ def remove_nil_fields
...
+        if @options[:'url-catalog'].blank?
+          @options.delete(:'url-catalog')
+        else
+          @options[:'url-catalog'] = @options[:'url-catalog'].join(',')
+        end
+      end
+    end
+
+    def prepare_access_token
+      if @options[:'access-token'].is_a?(AccessToken)
```
0robustus1: Why the check for the specific class? I would've assumed that it is only important if the value is non-nil.

lib/router_constraints.rb
```diff
@@ -130,9 +130,18 @@ def matches?(request, path = nil)
 
 class RefIRIRouterConstraint < IRIRouterConstraint
   def matches?(request)
+    regex = %r{
```
0robustus1: why not use a constant?

lib/router_constraints.rb
```diff
@@ -130,9 +130,18 @@ def matches?(request, path = nil)
...
+      |
+      \?([^=;&]+)[&;].*$ # splits a lone MMT-arg from normal query-string
+      |
+      \?.*$ # drops a the whole standard query-string
+    }x
     # remove the ref/:version_number portion from path
     path = Journey::Router::Utils.unescape_uri(request.original_fullpath).
-      sub(%r{\A/ref/\d+/}, '')
+      sub(%r{\A/ref/\d+/}, '').
+      sub(regex, '\1')
```
0robustus1: Is the MMT-check still current? (we have the `MMTRouterConstraint` now)

lib/hets/hets_options.rb
eugenk: The `prepare_*` methods convert the objects to `String` for passing them to the Hets API. This if condition checks if it already has been converted.

lib/router_constraints.rb
eugenk: You're right, it shouldn't be necessary anymore. Do we want to support access tokens with MMT-style URIs?

app/models/repository/access.rb
eugenk: An old habit from using Haskell: "Want to do something with all the elements? Use map." ;)
I will change this to `each`.

lib/hets/hets_options.rb
0robustus1: Yeah, but `to_s` is basically a noop on Strings, so you could just skip the altogether, couldn't you?

lib/router_constraints.rb
0robustus1: I'm not sure, to be honest. Access-Tokens are an internal feature (not exposed to the public), and we only supply `/ref/version_number`-based loc/ids to that. So is there any reason to support MMT?

eugenk: That's what I thought, too. So I just remove the MMT-handling code above and do nothing more about MMT.

0robustus1: Alright.

0robustus1:
:+1:

PR #1139, opened 2014-12-06
28 remote file upload

### LINKED ISSUE #28: No filename extension appended to extension-less ontology files?
Importing http://omdoc.org/ontology fails with Hets expecting CASL.

BTW note that the right URL is actually http://omdoc.org/ontology#, but this doesn't work at the moment; see #27.

I suspect the cause is that the downloaded file is named "ontology", without a filename extension.

But I thought we had said earlier that, for now, extension-less downloaded files should be renamed to *.owl, as in most cases what you get from such "cool URI" URLs is OWL (or RDF, which Hets does not yet support completely).

0robustus1:
Should fix #28. Needs the current version of hets.

### DIFF
--- a/app/assets/javascripts/files.js.coffee
+++ b/app/assets/javascripts/files.js.coffee
@@ -1,5 +1,10 @@
+filename_input = 'input#repository_file_target_filename'
+remote_file_input = 'input#repository_file_remote_file_iri'
+current_filename = null
+
 jQuery ->
   form_hooks()
+  remote_iri_hook()
 
 reverse_locality = (locality) ->
   if locality == 'remote'
@@ -10,6 +15,22 @@ reverse_locality = (locality) ->
 valid_locality = (locality) ->
   not not reverse_locality(locality)
 
+determine_filetype = (iri, callback) ->
+  payload = iri: iri
+  $.post "/filetypes/", payload, callback, 'json'
+
+determine_filename = (callback) ->
+  iri = $(remote_file_input).val()
+  if iri.length
+    result = iri.match(/\/([^/]+)(\?.*)?$/)
+    callback(result[1], iri) if result
+
+has_fileextension = (filename) ->
+  filename.split('.').length > 1
+
+filename_eligible = ->
+  $(filename_input).val().length == 0 or
+    current_filename == $(filename_input).val()
 
 form_hooks = ->
   return unless $('form#new_repository_file')
@@ -18,3 +39,12 @@ form_hooks = ->
     if valid_locality type
       $("#group-file_upload_#{type}").show()
       $("#group-file_upload_#{reverse_locality type}").hide()
+
+remote_iri_hook = ->
+  $(remote_file_input).on 'keyup', (e) ->
+    determine_filename (filename, iri) ->
+      if not has_fileextension(filename) and filename_eligible()
+        determine_filetype iri, (filetype) ->
+          suggested_filename = "#{filename}#{filetype.extension}"
+          $(filename_input).val(suggested_filename)
+          current_filename = suggested_filename
--- a/app/assets/javascripts/files.js.coffee
+++ b/app/assets/javascripts/files.js.coffee
@@ -1,0 +1,20 @@
+jQuery ->
+  form_hooks()
+
+reverse_locality = (locality) ->
+  if locality == 'remote'
+    'local'
+  else if locality == 'local'
+    'remote'
+
+valid_locality = (locality) ->
+  not not reverse_locality(locality)
+
+
+form_hooks = ->
+  return unless $('form#new_repository_file')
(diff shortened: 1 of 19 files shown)


app/models/repository_file.rb
```diff
@@ -56,7 +59,29 @@ def initialize(*args, &block)
...
+      @remote_file ||= retrieve(remote_file_iri)
+    end
+  end
+
+  def source_filename
+    case file_upload_type
+    when 'local'
+      temp_file.original_filename
+    when 'remote'
+      remote_file_iri.split('/').last
```
eugenk: What if the iri has a query string or ends on `/`?

app/assets/javascripts/files.js.coffee
```diff
@@ -0,0 +1,50 @@
...
+  not not reverse_locality(locality)
+
+determine_filetype = (iri, callback) ->
+  payload = iri: iri
+  $.post "/filetypes/", payload, callback, 'json'
+
+determine_filename = (callback) ->
+  iri = $(remote_file_input).val()
+  if iri.length
+    result = iri.match(/\/([^/]+)(\?.*)?$/)
```
eugenk: Here, you consider query strings.

db/migrate/20141206161905_create_file_extension_mime_type_mappings.rb
```diff
@@ -0,0 +1,10 @@
+class CreateFileExtensionMimeTypeMappings < ActiveRecord::Migration
+  def change
+    create_table :file_extension_mime_type_mappings do |t|
+      t.string :file_extension, null: false
+      t.string :mime_type, null: false
+
+      t.timestamps
+    end
```
eugenk: How about
```ruby
add_index :file_extension_mime_type_mappings, :mime_type
```
since you call `find_by_mime_type`?

db/migrate/20141206163145_add_mappings_to_file_extension_mime_type_mappings.rb
```diff
@@ -0,0 +1,28 @@
+class AddMappingsToFileExtensionMimeTypeMappings < ActiveRecord::Migration
+  MAPPINGS = {
+    '.clif' => 'text/clif',
+    '.owl' => 'application/owl',
+    '.owl' => 'text/obo',
```
eugenk: You use the key `'.owl'` twice.

app/controllers/filetypes_controller.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  protected
+
+  def filetype
+    @filetype ||= Hets::FiletypeCaller.new(HetsInstance.choose).
+      call(params[:iri])
+  end
+
+  def filetype_json
```
eugenk: Is it a convention to name a method `_json`, when the return hash is supposed to be rendered as json?

app/assets/javascripts/files.js.coffee
```diff
@@ -0,0 +1,50 @@
...
+    callback(result[1], iri) if result
+
+has_fileextension = (filename) ->
+  filename.split('.').length > 1
+
+filename_eligible = ->
```
eugenk: I prefer `filename_suggest_eligible`.

eugenk:
Entering the remote address `https://ontohub.org/dol-examples/equiv` into the form, I get the filename suggestion `equivnull`.

app/models/repository_file.rb
0robustus1: - ends on `/`: we don't care as `split` will take care of this (a lone `/` at the end will just be dropped)
- contains query-string: i will fix that.

app/controllers/filetypes_controller.rb
0robustus1: I believe i've seen this on a few projects. If it however is not a convention, it should now be an ontohub convention to name methods in such a way, don't you think?

db/migrate/20141206163145_add_mappings_to_file_extension_mime_type_mappings.rb
0robustus1: Oh yes. I've overlooked that. I change it to use nested lists.

0robustus1:
Mh i don't get the same issue. For me it suggests: `equiv.het`. Which isn't perfect either but it is what hets returns. Which version of hets are you using?

app/controllers/filetypes_controller.rb
eugenk: Alright.

eugenk:
When I wrote the comment, it said `text/plain`. Now it says `text/het` and the correct extension is suggested.
I don't know why it changed.

eugenk:
:+1:

PR #1144, opened 2014-12-15
Re add oops states test

eugenk:
built on top of #1077


0robustus1:
I rebased staging and tried my hand on fixing this thing.

eugenk:
stalled because we don't know why the test is red and when we want to tackle the problem.

PR #1147, opened 2014-12-23
Fix failing link spec due to hets error

eugenk:
This commit fixes the failing link_spec caused by https://github.com/spechub/Hets/issues/1433.
It allows for old and new hets versions to pass this test.

It is supposed to be reverted as soon as the hets issue is resolved.


eugenk:
The hets-issue is resolved. Does not need to be merged anymore.

PR #1148, opened 2014-12-23
Revert workaround hets issue 1433

eugenk:
This reverts the workaround of #1147. This is supposed to be merged as soon as https://github.com/spechub/Hets/issues/1433 is solved.


eugenk:
The hets-issue is resolved. Does not need to be merged anymore.

PR #1162, opened 2015-01-16
Add setting for hets workers count.

### LINKED ISSUE #1150: god doesn't reliable terminate workers
"god terminate" does a very poor job and always leaves sidekiq workers running often all-1|2 (leaks ~32x256MiB ~= 8GiB RAM RSS) even when they are all idle! 

Furthermore it would be nice, if one could configure, how many sidekiq workers get spawned by god. The current default (nrpoc+4?) seems to be inappropriate. My rule of thumb says, nproc*(0.75+-0.05) would be more healthy for a system.

eugenk:
This shall fix #1150.
Now the number of hets workers can be configured by the admin, but it is restricted to
* at least one
* at most "number of processors" minus "number of other high load workers" (2)

### DIFF
--- a/config/god/app.rb
+++ b/config/god/app.rb
@@ -3,12 +3,15 @@ require File.expand_path('../hets_workers',  __FILE__)
 
 RAILS_ROOT = ENV['RAILS_ROOT'] || File.dirname(__FILE__) + '/../..'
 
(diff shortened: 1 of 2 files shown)


0robustus1:
:+1:

PR #1164, opened 2015-01-16
Colorize the output, if the environment says so.

eugenk:
This enables colorized rspec output in jenkins.


eugenk:
The URL of the default check details is wrong now, as I renamed the job in jenkins. The correct URL is http://john.iws.cs.ovgu.de:49153/job/ontohub%20pull%20requests/4/

eugenk:
test again

ebolloff:
So beautiful :+1:

PR #1169, opened 2015-01-22
Improve ontology list display

0robustus1:
Should basically be as advertised: Improves spacing for ontology-fancy-link (the svg-icons) in specific lists. This should remove the overlap between multiple heterogeneous-ontology icons.


eugenk:
:+1:

PR #1173, opened 2015-01-29
1171 add commit model

### LINKED ISSUE #1171: Add git-commit header-fields to ontology-version
Currently the Ontology-Versions only stores a reference to the commit-oid. If we were however to also
store committer, commit-date, author and author-date information we would be able to perform
rather fast queries on that data.

I'm proposing this because i'm currently working on implementing the Date-Range feature of LocIds (e.g.
`'http://ontohub.dev/ref/2015-01-27/default/partial_order//strict_partial_order'`). And even if we were to get that working with rugged, it would still be quite slow (which isn't very good for HTTP-Routing).

@eugenk as you are currently the one person with the most experience in rugged, what would be your take on this?

0robustus1:
Should fix #1171, which again should help with the completion of #1154.

### DIFF
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -1,0 +1,7 @@
+class Commit < ActiveRecord::Base
+  attr_accessible :commit_oid, :repository
+  attr_accessible :author, :author_date
+  attr_accessible :commit_date, :committer
+
+  belongs_to :repository
+end
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -1,7 +1,7 @@
 class Commit < ActiveRecord::Base
   attr_accessible :commit_oid, :repository
   attr_accessible :author, :author_date
-  attr_accessible :commit_date, :committer
+  attr_accessible :committer, :commit_date
 
   has_many :ontology_versions
   has_many :ontologies, through: :ontology_versions
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -3,5 +3,8 @@ class Commit < ActiveRecord::Base
   attr_accessible :author, :author_date
   attr_accessible :commit_date, :committer
 
+  has_many :ontology_versions
+  has_many :ontologies, through: :ontology_versions
+
   belongs_to :repository
 end
--- a/app/models/commit.rb
+++ b/app/models/commit.rb
@@ -7,4 +7,15 @@ class Commit < ActiveRecord::Base
   has_many :ontologies, through: :ontology_versions
 
   belongs_to :repository
+
+  def fill_commit_instance!
+    commit = repository.git.repo.lookup(commit_oid)
+    data = commit.committer
+    self.committer = "#{data[:name]} <#{data[:email]}>"
+    self.commit_date = data[:time]
+    data = commit.author
+    self.author = "#{data[:name]} <#{data[:email]}>"
+    self.author_date = data[:time]
+    save!
+  end
 end
--- a/app/models/repository.rb
+++ b/app/models/repository.rb
@@ -13,6 +13,7 @@ class Repository < ActiveRecord::Base
 
   has_many :ontologies, dependent: :destroy
   has_many :url_maps, dependent: :destroy
+  has_many :commits, dependent: :destroy
 
   attr_accessible :name,
                   :description,
--- a/app/models/repository/git_repositories.rb
+++ b/app/models/repository/git_repositories.rb
@@ -117,6 +117,7 @@ module Repository::GitRepositories
     version = ontology.versions.build(
       { commit_oid: commit_oid,
         user: ontology_version_options.user,
+        commit: commit_for!(commit_oid),
         # We can't use the ontology's filepath bacause it might have changed
         basepath: File.basepath(ontology_version_options.filepath),
         file_extension: File.extname(ontology_version_options.filepath),
@@ -126,7 +127,6 @@ module Repository::GitRepositories
     version.save!
     ontology.ontology_version = version
     ontology.save!
-    commit_for!(commit_oid)
 
     version
   end
--- a/app/models/repository/git_repositories.rb
+++ b/app/models/repository/git_repositories.rb
@@ -138,16 +138,7 @@ module Repository::GitRepositories
   def commit_for!(commit_oid)
     instance = Commit.where(repository_id: self, commit_oid: commit_oid).
       first_or_initialize
-    if !instance.persisted?
-      commit = git.repo.lookup(commit_oid)
-      data = commit.committer
-      instance.committer = "#{data[:name]} <#{data[:email]}>"
-      instance.commit_date = data[:time]
-      data = commit.author
-      instance.author = "#{data[:name]} <#{data[:email]}>"
-      instance.author_date = data[:time]
-      instance.save!
-    end
+    instance.fill_commit_instance! unless instance.persisted?
     instance
   end
 
--- a/app/models/repository/git_repositories.rb
+++ b/app/models/repository/git_repositories.rb
@@ -34,7 +34,8 @@ module Repository::GitRepositories
   end
 
   def delete_file(filepath, user, message = nil, &block)
-    git.delete_file(user_info(user), filepath, &block)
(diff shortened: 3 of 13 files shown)


db/migrate/20150128133702_create_commits.rb
```diff
@@ -0,0 +1,19 @@
+class CreateCommits < ActiveRecord::Migration
+  def change
```
0robustus1: Don't care, as it is in a migration.

app/models/repository/git_repositories.rb
```diff
@@ -128,6 +131,22 @@ def create_version(ontology, commit_oid, ontology_version_options)
     version
   end
 
+  def commit_for!(commit_oid)
+    instance = Commit.where(repository_id: self, commit_oid: commit_oid).
+      first_or_initialize
+    if !instance.persisted?
```
0robustus1: With the long body i prefer `if`. @eugenk, what do you say?

app/models/repository/git_repositories.rb
```diff
@@ -128,6 +131,22 @@ def create_version(ontology, commit_oid, ontology_version_options)
     version
   end
 
+  def commit_for!(commit_oid)
```
0robustus1: Can't really shorten it. Do you see a good way @eugenk?

app/models/commit.rb
```diff
@@ -0,0 +1,10 @@
+class Commit < ActiveRecord::Base
+  attr_accessible :commit_oid, :repository
+  attr_accessible :author, :author_date
+  attr_accessible :commit_date, :committer
```
eugenk: `attr_accessible :committer, :commit_date` for consistency in the ordering with the author-line.

app/models/commit.rb
```diff
@@ -0,0 +1,10 @@
...
+
+  has_many :ontology_versions
+  has_many :ontologies, through: :ontology_versions
+
+  belongs_to :repository
+end
```
eugenk: Just a note: You didn't include the commit message into the commit model, which is the only interesting thing a commit has left. However, I don't think we need this information in the database right now.

app/models/repository/git_repositories.rb
eugenk: How about something like
```ruby
  def commit_for!(commit_oid)
    instance = Commit.where(repository_id: self, commit_oid: commit_oid).
      first_or_initialize
    setup_commit_instance(instance) unless instance.persisted?
    instance
  end

  def setup_commit_instance(instance)
    commit = git.repo.lookup(instance.commit_oid)
    data = commit.committer
    instance.committer = "#{data[:name]} <#{data[:email]}>"
    instance.commit_date = data[:time]
    data = commit.author
    instance.author = "#{data[:name]} <#{data[:email]}>"
    instance.author_date = data[:time]
    instance.save!
  end
```

app/models/repository/git_repositories.rb
eugenk: We agreed to use `if` if the body spans over more than three lines, so this is okay. However, see the suggestion above.

db/migrate/20150128133702_create_commits.rb
eugenk: Agree.

app/models/repository.rb
```diff
@@ -13,6 +13,7 @@ class Repository < ActiveRecord::Base
 
   has_many :ontologies, dependent: :destroy
   has_many :url_maps, dependent: :destroy
+  has_many :commits, dependent: :destroy
```
eugenk: Name clashes with https://github.com/ontohub/ontohub/blob/staging/app/models/repository/git_repositories.rb#L6

lib/tasks/migrate_data.rake
```diff
@@ -0,0 +1,17 @@
+def create_commits!(repository)
+  repository.git.commits { |c| repository.commit_for!(c.oid) }
+end
+
+def attach_commit!(ontology_version)
+  commit = Commit.find_by_commit_oid(ontology_version.commit_oid)
```
eugenk: I know, it's very improbable, but how about
```ruby
  commit = ontology_version.repository.commits.
    find_by_commit_oid(ontology_version.commit_oid)
```
with `repository.commits` meaning the association to the commit model and not the commits-iterator.

spec/models/commit_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'spec_helper'
+
+describe Commit do
+  pending "add some examples to (or delete) #{__FILE__}"
```
eugenk: We could have tests for the associations, so the file has the right to exist:
```ruby
  context 'associations' do
    %i(ontology_versions ontologies).each do |association|
      it { should have_many(association) }
    end

    it { should belong_to(:repository) }
  end
```

spec/models/repository_directory_spec.rb
```diff
@@ -26,30 +26,42 @@
...
-        repository.save_file_only(Tempfile.new('tempfile').path, filepath, 'message', user)
-      end
 
-      it 'should fail when path is the directory' do
-        repository_directory.name = dirname
-        expect(repository_directory.valid?).to be_falsy
+      let(:save_operation) do
+        file = Tempfile.new('tempfile').path
+        repository.save_file_only(file, filepath, 'message', user)
       end
```
eugenk: This is a funny diff highlighting. Only `do`, `end` and `=` are the same. Nice, github! :D

app/models/commit.rb
0robustus1: Really? The only thing? I could think of a few. However i did not include them, as we do not need them right now...

e.g. commit-message, files affected by commit, the diff of the commit, the way files were affected (added, deleted, changed, rewritten).

app/models/repository.rb
0robustus1: Ah, that was the problem i could not find...

lib/tasks/migrate_data.rake
0robustus1: Actually i meant to do that, don't know why i didn't.

spec/models/commit_spec.rb
0robustus1: mh, alright.

lib/tasks/migrate_data.rake
```diff
@@ -0,0 +1,18 @@
+def create_commits!(repository)
+  repository.git.commits { |c| repository.commit_for!(c.oid) }
+end
+
+def attach_commit!(ontology_version)
+  commit = ontology_version.repository.commits.
+    find_by_commit_oid(ontology_version.commit_oid)
```
jelmd: Just wondering, why the dot is at the end of the line instead of right at the beginning of the called object method?
Without it, it looks like a "static"/"target unrelated"/"dangling" call - IMHO a little bit confusing/harder to read (at least in Java it is a common coding rule to start a line continuation with a dot after the indent if possible - not sure, whether ruby allows that at all).

0robustus1: We end on line-continuation with `.` as it is in our style guide. It is in our style guide because `.` at the end of the line instead of the beginning allows for comments between those two lines. When placing the `.` at the beginning of the line comments would result in syntax errors.

jelmd: Ah ok, too bad. Sounds like being in python ;-)

0robustus1: And to be honest i actually find it easier to read. This way the reader knows, when reaching the end of the line, that the call-chaining is not over instead of being surprised later on.

jelmd: Hmm, ok. Basically the reason for .@SOL is how the brain of ppl usually works, i.e. reading is usually: find the anchor (start) and than use pattern matching and stop if it appears, that we got it/can ignore the rest. So most of the time  "reading" doesn't actually switches on the region, which really _parses_ the text and stops much earlier wrt. end of line than we think ;-) - otherwise it would be too in-efficient, costs too much resources/power aka concentration and we would need a lot more good food ;-) . Anyway, with some training ... ;-)

eugenk:
:+1:

PR #1176, opened 2015-02-04
add Puma support

jelmd:
This patch adds out-of-the-box puma support. Puma is a simple, fast, and highly concurrent HTTP 1.1 server for Ruby web applications.

'til now most setups use the [Phusion Passenger] (https://github.com/phusion/passenger) Apache httpd plugin to multiplex/route related HTTP[S] requests to the ontohub web app. Unfortunately this doesn't work in LCX containers because of internal passenger bugs. 2 month after submission of a [bug report] (https://github.com/phusion/passenger/issues/1315) there is still not a single comment from the Phusion team (and BTW, even the mentioned workaround found doesn't fix the problem to 100%).  Looking at the related [Issue page] (https://github.com/phusion/passenger/issues) suggests, that this SW has much more problems/bugs, as their maintainers are willingly to admit (and able to fix in time?). Furthermore Passenger is a pretty heavy-weight application (written in C++ - so no wonder...) for ruby, python and node.js - ontohub web app is just ruby.  Last but not least it seems, that the use of the non-commercial Passenger plugin doesn't buy us anything, except trouble and higher complexity wrt. to the env. setup/troubleshooting.

So with this patch it is recommended to drop passenger and to use puma instead. Puma is a very light-weight ruby application and thus fits much better into the ontohub environment than passenger. No bugs have been discovered from my side so far and small [faban] (http://faban.org) based microbenchmarks wrt. http://ontohub.org/ show, that it is even faster than the Passenger. Since puma is a normal web server, running an Apache httpd infront of it, is as easy as e.g. adding `ProxyPass / http://localhost:3000/` to its apache config - so no need to learn special things (like passenger peculiarities) and one has still the same flexibility (actually even a little bit more) as with passenger.

Since puma sits on top of the web application, no changes wrt. the onthub app itself are required, if one chooses to use puma instead of Passenger.


config/puma.rb
```diff
@@ -0,0 +1,36 @@
...
+# We don't like noise
+quiet
+# IMPORTANT wrt. MRI ruby - saves rsources!
+preload_app!
+
+# let's make sure, that ppl with the same GID as the running puma process
+# are really able to control puma (the group needs write access). See option
+# --control and --control-url above.
+on_worker_boot do
+	str = @options[:control_url]
```
eugenk: I prefer a more expressive name:
```ruby
control_url = @options[:control_url]
```

eugenk:
Please run `bundle install` and include the `Gemfile.lock` in this branch.

Gemfile
```diff
@@ -146,6 +146,9 @@ end
 group :production do
   gem 'god', '~> 0.13.4'
   gem 'exception_notification', '~> 4.0.1'
+  # a light-weight alternative for the heavy-weight, buggy Phusion apache httpd
+  # plugin. See https://github.com/puma/puma , config/puma.rb
+  gem 'puma', '>= 2.10.0'
```
eugenk: Please use `~>` instead of `>=`. This way we most likely won't have trouble with incompatibilities introduced by newer versions.

jelmd:
Disagree - '>=' was choosen for good reasons. If I didn't understand it wrong, the dependency is actually: choose a puma version compatible to the rack API used by the web application and which in turn - IIUC - is implicitly defined by `gem 'rails', '~> 3.2.19'`. Therefore specifying any version for puma results IMHO into an overconstrained dep, which just makes live harder for no reason. So the intention for `>= 2.10.0` is basicallly document the first version it should definitely work with (we started working with it) and implicitly a "don't annoy me with your useless messages" wrt. the not-so-smart tools.

config/puma.rb
jelmd: Well, since I did not debug the application and nowhere is documented, what it really contains, at this point it is for me just a string. Using misleading names like control_url may lead other developers to assume, that its value has already been validated, which in turn may cause sooner or later a SW fauxpas like with the [Mars Climate Orbiter] (http://en.wikipedia.org/wiki/Mars_Climate_Orbiter#Cause_of_failure) ;-) .

eugenk:
Basically you are right, but there can still be breaking changes wrt. to our whole setup. Every time we update gems, we need to test if everything is still working. This is why we only update gems manually (except for security patches).

We had a case once (rugged), that was marked as compatible *with every other gem* but still introduced incompatibilities *with our code* which made a lot of tests fail. The search for the reason was not very pleasant. ;)

config/puma.rb
eugenk: You are right, the documentation is not very extensive. https://github.com/puma/puma/blob/master/lib/puma/control_cli.rb#L41 is the only piece of documentation I could find about it. According to that line of code you can assume that the string is really the control url, because it's passed to puma as an argument by the user.

eugenk: Oh, the value `"auto"` is accepted as well, but in my humble opinion it's still valid to use the name `control_url`.

jelmd: The given URL is obviously unrelated and doesn't proof, that it contains "more" than just a string. Anyway, it is a riddle to me, why we should waste our valuable time with such "out-of-the-scope" stuff: If there is no API, one can't rely on anything unless the complete application got analyzed and one can be 100% certain, that its assumption is right - sorry, but I've no time for such nonsense!  When a new version comes in, one has to do it again - sorry again, I've no time for such nonsense.  And last but not least and already said, the web application is using rails and thus this is the only concern one has to deal with. It is beyond the scope of an web app using a certain framework to dictate, which exact web server should be used, and AFAICS for passenger I can't see any similar constraints. And yes, strictly taken, this doesn't belong into the web app, it just provides out-of-the-box support (I didn't make the rules, how ruby apps get assembled/packaged ...). So sink or swim, take or discard - these are my last words on this - already wasted too much time on this.

eugenk:
I'm closing this pull request in favor of #1241.

PR #1195, opened 2015-02-11
1190 restructure migrations

### LINKED ISSUE #1190: Fix migrations
Fix migrations by
* Putting all data-consistency-migrations, which need to be run on ontohub.org and staging.ontohub.org, into rake tasks. Also create a list of rake tasks which need to be run on deploy (containing rake task and the corresponding date/timestamp).
* Squashing all migrations, that are already on ontohub.org, into one migration.

eugenk:
This shall fix #1190.

* Used [squasher](https://github.com/jalkoby/squasher) to squash the migrations prior to the last Ontohub release into one big schema.
* Adds [data_migrate](https://github.com/ajvargo/data-migrate) to separate schema migrations from data migrations. The important data like ontology file extensions and file extension to mime type mappings are still in the schema migrations because we always want to migrate them instantly and they don't depend on other Ontohub code.

As for this pull request, the data migrations can be executed with `rake data:migrate`. This is a synchronous job. It can be executed asynchronously in two ways:
* in an ssh connection with `nohup rake data:migrate &`. Then the job continues its work even if the ssh connection is closed. Its output is written into a file called `nohup.out`.
* by `rake data:migrate:async` which creates a Sidekiq job that runs `rake data:migrate`. Interestingly, on my machine, the job is not added to the `default` queue, but to the queue with the empty name and stay "Enqueued". I can't explain why this happens. Your ideas are welcome :)

For future coding: Other than schema migrations, data migrations are supposed to be changed whenever the code they depend on changes.


eugenk:
Sorry Hound, I ignore your comments on the 20140615175553_init_schema.rb because that file is mostly generated.

eugenk:
test this please

0robustus1:
:+1:

Issue #1170, opened 2015-01-27
Integrate parts of server config into Ontohub


tillmo:
see https://gist.github.com/tillmo/a7cb0537bd41872b411a, where the diff is commented. Some diffs should be turned into branches (forking from staging), others should be avoided.
Moreover, here are some further hints for the server config:
* install dependencies: svn, git-svn
* entities errors: stem from migration executed in production mode
* switch to production mode; use symlink for hets instead of source patches

jelmd:
After applying the follwoing patches, all customization (if needed at all now) can be done via config/settings.[local].yml and friends, so no patching of other files is required anymore:
```
  1170-frontend-notes
  1170-git-data_dir
  1170-hets-binary-path
  1170-puma-support
  1170-settings-cleanup
  common-settings-behavior
```
The latest https://github.com/ontohub/admin-scripts/tree/puma/scripts/install expect these patches to be installed or available as ~ontohub/etc/*.patch.

eugenk:
This is not completed yet. Reopening.

eugenk:
The pull requests #1241, #1266, 1268, #1269 and #1309 are part of this issue.

tillmo:
we have decided to add puma to the Gemfile, as in @jelmd's pull request #1176. @eugenk, could you please create an appropriate pull request? (The puma config has already been added in #1241).

PR #1197, opened 2015-02-11
1170 git data dir

jelmd:
see https://github.com/ontohub/ontohub/issues/1170

### DIFF
--- a/app/models/key.rb
+++ b/app/models/key.rb
@@ -24,10 +24,12 @@ class Key < ActiveRecord::Base
 
   def add_to_authorized_keys
     AuthorizedKeysManager.add(self.id, self.key)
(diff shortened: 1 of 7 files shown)


eugenk:
Instead of making those huge amends, you could set
```ruby
  config.data_root = '/data'
```
in the `config/environment/production.local.rb` This file overrides all previously set settings (i.e. also application.rb).

Then, *all the data* will be stored in `/data`, i.e. not only the bare git repositories but also the symlinks from repository path identifiers to the bare repository and checked out files for download/analysis etc.

jelmd:
Despite it would be the wrong place blame juliank for putting it there, not me.
Wrt. to commit message blame github , not me.

eugenk:
I'm closing this pull request in favor of #1268.

Issue #1200, opened 2015-02-12
call Hets with auto command


tillmo:
see https://github.com/spechub/Hets/wiki/RESTful-Interface, `GET /dg/coded-iri/command-list?options`. The command-list should include `auto`. The reason is that then the development graph calculus will be automatically applied, and all necesary proof obligations be generated for the individual ontologies.

0robustus1:
Closed with #1201.

PR #1201, opened 2015-02-12
1200 call hets with auto

0robustus1:
Shall solve #1200.

### DIFF
--- a/config/god/hets_workers.rb
+++ b/config/god/hets_workers.rb
@@ -5,14 +5,14 @@ class HetsWorkers < Watcher
     'hets'
   end
 
-  def start_cmd(hets_opts = [])
-    hets_opts =
-      if hets_opts && !hets_opts.empty?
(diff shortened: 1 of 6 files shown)


tillmo:
wow, that was quick. While we are changing the options: could you also please add the option `--casl-amalg=none`? This prevents Hets from a time-consuming analysis of architectural specifications, which we (currently) do not need for Ontohub anyway.

0robustus1:
Should be done with a6ef1637eefa4be07a8faf1abcb51a038e072d0c for `hets -X`, for cmd-line tasks we were already using `-a none` which is actually the short version of `--casl-amalg=none`.

eugenk:
:+1:, but I prefer `--casl-amalg=none` over `-a none` and `--apply-automatic-rule` over `-A`. If you change it or not, is in your hands. You get the sign off either way.

PR #1202, opened 2015-02-13
1061 handle hets api errors

### LINKED ISSUE #1061: Handling of hets-errors in API
We should handle hets-errors at least just as well in the API as we did when performing system calls.

0robustus1:
Shall partially fix #1061. It provides basic error handling in which the user will actually get the
hets-error message where possible. It however probably requires improvement, which i will do in a later pull-request. But seeing as this seems to be an important issue, we should still review and merge this pull-request.


eugenk:
:+1:

PR #1203, opened 2015-02-13
frontend related cleanup

jelmd:
- public/.htaccess is frontent specific (apache httpd, only) and thus
  is out of the application scope, belongs as a side note to docs/wiki
- see https://github.com/ontohub/ontohub/issues/1170


eugenk:
Closing this in favor of #1270.

PR #1206, opened 2015-02-16
make settings behavior consistent

jelmd:
- introduces AppConfig module, which allows non-Rails apps to use the same
  Settings the same way as RailsConfig enhanced Rails::Application do
  (i.e. same order AND the _same_ set of files!)
- direct/inconsistent handling of single yml files not needed anymore
- see lib/environment_light.rb for examples
- minor fixes of wrong settings


tillmo:
This pull request breaks the git clone functionality. #1231 shall fix this.

eugenk:
Closing this in favor of #1236 and #1269.

PR #1194, opened 2015-02-10
1154 support ref locids

0robustus1:
This branch/Pull-Request is based on 1154-add_loc-ids/#1179 instead of on staging.
The reason for this is the fact that it needs the previous loc/id work. As soon as the other pull-request
is merged, i will rebase on staging.

This Pull-Request completes the second half and shall therefore close #1154.

PR #1207, opened 2015-02-16
Fix renaming of link to mapping

eugenk:


app/helpers/link_helper.rb
```diff
@@ -94,19 +94,19 @@ def ontology_link_to(resource)
     end
   end
 
-  def sort_link_list(collection)
+  def sort_mapping_list(collection)
```
eugenk: I think 11 is okay.

eugenk:
I see it's already fixed in 919ddcc410e2ed682a9691187bbf58132186e543 of #1194.

PR #1208, opened 2015-02-16
1145 regroup concerns

### LINKED ISSUE #1145: Repository validations should be in a single file
Currently, validations of `Repository` are spread across several files. They all should be in the validations concern.

eugenk:
This shall fix #1145 and some other issues that bothered us for a long time.
We had some concerns that were not grouped sensibly. I pushed around some of the code to put a better structure in the concerns of `Repository` and `Ontology`. There is no new code in this branch - just old code moved around a little.

### DIFF
--- a/app/models/ontology.rb
+++ b/app/models/ontology.rb
@@ -7,6 +7,7 @@ class Ontology < ActiveRecord::Base
   # Ontology Model Includes
   include Ontology::Import
   include Ontology::Scopes
+  include Ontology::Validations
   include Ontology::States
   include Ontology::Versions
   include Ontology::Symbols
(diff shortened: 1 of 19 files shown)


app/models/ontology.rb
```diff
@@ -5,46 +5,29 @@ class Ontology < ActiveRecord::Base
   include Metadatable
 
   # Ontology Model Includes
+  include GraphStructures::SpecificFetchers::Mappings
+  include IRIUrlBuilder::Includeable
+  include Ontology::Associations
```
0robustus1: I would've thought that associations and attributes belong together, seeing as they together form the structure of the model.

0robustus1:
Seems like a good start, although there still some stuff about the concerns we could optimize: :+1:

PR #1214, opened 2015-02-17
1211 fix ref locids for subdirectories

### LINKED ISSUE #1211: loc_ids do not work with subdirectories
see http://develop.ontohub.org/casl-um/new/Chapter3///ontology_versions
the url http://develop.ontohub.org/ref/1/casl-um/Chapter3?path=new is generated and used by hets (the suffix .env is added by hets after this url fails)

0robustus1:
Shall fix #1211. Additionally provides the migration to update existing data with locids.

### DIFF
--- a/spec/models/ontology_version_spec.rb
+++ b/spec/models/ontology_version_spec.rb
@@ -34,6 +34,25 @@ describe OntologyVersion do
       stub_hets_for(hets_out_file('pizza'))
     end
 
+    context 'in subdirectory' do
+      let(:ontology) { create :ontology, basepath: 'subdir/pizza' }
+      let(:qualified_locid) do
+        "http://#{Settings.hostname}#{ontology_version.locid}"
+      end
+
+      before do
+        ontology_version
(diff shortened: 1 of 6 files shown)


spec/models/ontology_version_spec.rb
```diff
@@ -34,6 +34,26 @@
...
+
+
+      before do
+        ontology_version
+        allow_any_instance_of(Hets::ParseCaller).to receive(:call) do |iri, *args|
+          throw(:iri, iri)
+        end
+      end
+
+      it 'should use the locid-ref for calling the parse-caller' do
```
0robustus1: Who reactived preference of double-quotes?

eugenk: Seems to be a bug in hound/rubocop because the yml says:
```yml
Style/StringLiterals:
  Description: 'Checks if uses of quotes match the configured preference.'
  Enabled: true
  Style: 'single_quotes'
  EnforcedStyle: single_quotes
```

eugenk:
Travis fails because of https://github.com/spechub/Hets/issues/1455. This can be merged :+1:

PR #1215, opened 2015-02-18
Add stylecheck script

eugenk:
This script runs rubocop with a suitable config generated from the
`.hound.yml`. The hound config includes a key that's not compatible with
rubocop. The config is copied to a tempfile without that key.

#### Usage
You can check your code for compliance with the style guide by running
```
script/stylecheck
```
in the ontohub directory. You can also specify a list of files and folders to go
through with
```
script/stylecheck app/models/ontology.rb app/models/ontology/
```

### DIFF
--- a/script/stylecheck
+++ b/script/stylecheck
@@ -1,0 +1,19 @@
+#!/usr/bin/env ruby
+require 'tempfile'
+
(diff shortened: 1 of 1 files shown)


0robustus1:
:+1:

PR #1188, opened 2015-02-08
1131 prove theorems

eugenk:
This branch shall fix #1131. It allows ontohub to prove theorems of an ontology by pressing a button in the theorems view.

The views themselves are simplistic and not very beautiful, but they contain all the information we have. They will be re-made in another issue.

The tactic script is currently saved as JSON. I don't like this solution, but since there are users who want to try the proving feature, I wanted to bring it as fast as possible to develop.ontohub.org. In another issue, I will attack the representation of the tactic script.

Note for the reviewer:
I recommend reading this pull request commit by commit. This approach should help understanding the changes.

Files appearing to be new in `lib/hets/dg/` aren't actually new. I moved them there from `lib/hets/` and put them in a deeper nested module to distinguish them from the files in `lib/hets/prove/`.

Also, there will be conflicts with #1136 with the `{Get,Prove,Action}Caller` because that one introduces `Hets::Options`.

PR #1216, opened 2015-02-18
Disable rather annoying style checks.

eugenk:
Those two style checks are not according to our style guide. They annoyed me in #1188 and #1208.


0robustus1:
:+1:

PR #1221, opened 2015-02-20
Settings cleanup

jelmd:
- make mail settings work
- use reasonable defaults for most settings required to exists or which are
  sensible wrt. to the running RAILS_ENV
- validate settings were needed/makes sense
- makes all mail delibvery_methods work
- => with this change ontohub should run out of the box, i.e.
  config/application.rb ensures, that if a certain setting is needed, it
  verifies and if needed replaces/adds it with a _reasonable_ default.
  However, an appriopriate example/default values should still be documented
  within config/settings.yml, so that the user is able to fine tune its
  instance as needed w/o scratching its head ...


eugenk:
I noticed that you want to merge into the branch `develop` in all your pull requests and that the base of your branches is `develop`. `develop` is sort of our playground. Whenever we feel that a feature branch is ready and we want to try this new feature in combination with other new features, we merge them into develop and play around. This way, a lot of "bad code" comes into develop which never makes its way into a release.

The correct base for new branches is `staging`. This one is for finalized and stable features. There are by far less errors than in `develop`. Also, `staging` is the desired destination for pull requests.

eugenk:
Please don't change the app ontohub in this extent. We're fine with non-developers adjusting a few lines in order to make things easier or sort out bugs. Your pull requests are way more than just adjusting minor issues, though. There are pull requests like this one that redefine principles of the app and don't go the Rails way. It shows that you are not a rails (or plain ruby) developer. This project is very complex and if we ever want to get new developers attracted to contributing, we have to comply to the rails way and the ruby style guide. 

With changing this much code you make yourself a lot of work that's not going to be merged directly. I think it would be easier and save you some work if you created issues describing your problem in detail and maybe presenting an idea of a solution, but not a complete implementation of what you think is a good solution.

jelmd:
Yes, your are right, I actually you should have done that!

eugenk:
Closing this in favor of #1236.

PR #1222, opened 2015-02-20
1218 ontology state feature test

### LINKED ISSUE #1218: Do what is to do
Step one: Look at https://github.com/ontohub/ontohub/blob/f7670322c267c298084c3b40ed196cdf9a1f2bc7/features/step_definitions/ontology_state_steps.rb#L8 

Step two: Do what is to do.

0robustus1:
Fix #1218.


ebolloff:
I don't know, if we want to wait for hets, but :+1:

PR #1227, opened 2015-02-23
Add deploy task for data migrations.

eugenk:
Finalizes #1190. This adds a sidekiq job on deployment to migrate data.
Currently, we want to run this task manually. As soon as it looks stable, we can merge this into staging.


tillmo:
does it look stable now?

eugenk:
We don't have any new data migrations, so we can't tell if it's stable yet.

eugenk:
Oh and I forgot: We don't use capistrano at the moment, so this task is irrelevant for our current server setup.

PR #1236, opened 2015-03-02
1230 validate settings

### LINKED ISSUE #1230: Validate Settings
Define a class that validates the settings and call it in the initializers.

### LINKED ISSUE #1231: switch to rails_config 0.5.0
@jelmd needs this for his work on settings. However, switching to rails_config 0.5.0 seems to cause problems in lib/environment_light.rb. Can this be fixed?

eugenk:
This shall fix #1230 and it also shall fix #1231.

About #1230:
I added a SettingsValidator class which has a lot of lines of code, but is not very complex. The most lines come from specifying the validations to apply.

About #1231:
I forked rails_config to remove the dependency of Rails and to allow arrays to be overwritten (see commit message of 22b0b434f6a92b32c4c9364c11d625f34af67dd9).

Other things I have done:
I removed the sunspot config, because we switched from solr to elasticsearch.


lib/environment_light.rb
```diff
@@ -11,5 +11,12 @@
 require 'bundler/setup'
 
 # Load application settings
-require File.expand_path("../../config/initializers/rails_config", __FILE__)
-Settings = RailsConfig.load_files ["settings.yml","settings.local.yml", "settings/#{Rails.env}.yml"].map{|f| "#{Rails.root}/config/#{f}" }
+require Rails.root.join('config/initializers/rails_config.rb')
+Settings = RailsConfig.load_files([
+    'settings.yml',
```
eugenk: No thanks. In this case all the other options look bad.

lib/settings_validator.rb
```diff
@@ -0,0 +1,317 @@
...
+  def validate_specials
+    validate_hets_path
+  end
+
+  def validate_hets_path
+    key_chain = %i(yml hets_path)
+    hets_paths = get_value(key_chain)
+    unless hets_paths.any? { |path| File.executable?(path) }
+      @errors << ResourceNotFound.new(key_chain,
+        "Executable 'hets' not found in any of the given paths: #{hets_paths}")
```
eugenk: No thanks. In this case all the other options look bad.

lib/settings_validator.rb
```diff
@@ -0,0 +1,317 @@
...
+      @errors << NotADirectory.new(cfg_key_chain,
+        "Path is not a directory: #{path}")
+    end
+  end
+
+  def validate_directory_exists(key_chain)
+    path = get_value(cfg_key_chain)
+    unless File.directory?(path)
+      @errors << NotADirectory.new(cfg_key_chain,
+        "Path is not a directory: #{path}")
```
eugenk: No thanks. In this case all the other options look bad.

lib/settings_validator.rb
```diff
@@ -0,0 +1,317 @@
...
+    end
+  end
+
+  def validate_with(key_chain, &block)
+    if value = get_value(key_chain)
+      block.call(value)
+    end
+  end
+
+  def validate_directory_exists(key_chain)
```
eugenk: This is fixed.

lib/environment_light.rb
```diff
@@ -11,5 +11,13 @@
...
+    "settings/#{Rails.env}.yml",
+    "environments/#{Rails.env}.yml",
+    'settings.local.yml',
+    "settings.#{Rails.env}.local.yml",
+    "environments/#{Rails.env}.local.yml",
+    'hets.yml'
+  ].map { |f| Rails.root.join('config', f) })
```
eugenk: No thanks. In this case all the other options look bad.

0robustus1: How about the following:

```ruby
require Rails.root.join('config/initializers/rails_config.rb')
settings_files = %W(
  settings.yml
  settings/#{Rails.env}.yml
  environments/#{Rails.env}.yml
  settings.local.yml
  settings.#{Rails.env}.local.yml
  environments/#{Rails.env}.local.yml
  hets.yml
)
abs_settings_files = settings_files.map { |f| Rails.root.join('config', f) }
Settings = RailsConfig.load_files(abs_settings_files)

```

config/application.rb
```diff
@@ -76,6 +76,9 @@ class Application < Rails::Application
     config.external_url_mapping = APP_CONFIG = YAML.load(File.read(Rails.root + "config/external_link_mapping.yml"))
 
     config.before_initialize do
+      Settings.add_source!(Rails.root.join('config', 'hets.yml').to_s)
```
jelmd: Doesn't work because it doesn't allow the user to overwrite anything.

lib/environment_light.rb
jelmd: :-1:
This is "Repeat yourself" and the root cause inconsistencies: One should use the internally defined set of files (whatever this is) and prepend or append the other stuff. I already mentioned, why append is wrong above. BTW: Wondering, why hets.yml now gets always included? IIRC , e.g. the git stuff doesn't need it at all ...

config/application.rb
```diff
@@ -76,6 +76,9 @@ class Application < Rails::Application
...
+      Settings.reload!
+
```
jelmd: Why reload? If you always wanna include hets.yml (for whatever reason), it should be made clear  by adding a corresponding 'include' directive in settings.yml.

config/application.rb
```diff
@@ -87,5 +90,9 @@ class Application < Rails::Application
       end
 
     end
+
+    config.after_initialize do
```
jelmd: Running a settings validator after all initializers have been run is probably too late for most stuff/fruitless art. The initializers usually need to know, how they should initialize aka configure XY.

lib/settings_validator.rb
jelmd: Huh, do you want introduce a new project for meta data validation?
- where is the place to put the defaults?
- how can ranges for certain values be described (e.g. min max)
- how can be system properties access? E.g. to define a reasonable default like #{numcpus} * 0.75
- which system properties exist?
- how can be specified: use this in 'production' mode, use that in 'development', ... mode?

IMHO this is a nice try, but absolute over-engineering (one could use xml + a proper schema and existing schema validator and is through with it, perhaps even rng would be sufficient) and only gives one a more or less generic type checker. The real work is still missing, i.e. no application specific checks and no reasonable defaults and even the burden to actually set all the stuff (I deduce it from the 'HAS_TO_BE_PRESENT'), which in turn keeps the application from being smart (Most human beings would say: even my stupid ancient handy is able  to determine e.g. its hostname, and  such a "intelligent" app on a big iron can't do that - what's wrong with it?).

So, forking a different gem, writing about 300 lines of code, which doesn't provide, what is actually needed, introducing other errors/potential inconsistencies is for my taste a little bit too much. I still prefer "keep it small and simple", and thus for those few settings a simple check at a central place like already shown in the1170-settings-cleanup.patch would have saved you probably a lot of time. It's a pity wrt. to all the work you've done :(. Sometimes simple stuff is also the most efficient stuff, wrt. the code as well as the writer ;-)

There are probably other things I've missed, but this github stuff is such a pain in the ... with all its disturbing stuff between the lines, which makes proper reviews impossible ...

config/application.rb
eugenk: As far as I know, include is not possible in yml.

config/application.rb
eugenk: Well the important thing is that the application doesn't start when there are wring settings. And some of the initializers are for configuration and need to be validated as well. So I use `after_initialize` to catch those initializers as well.

lib/environment_light.rb
eugenk: True. I will look into this tomorrow.

lib/settings_validator.rb
eugenk: What do you mean by meta data validation?

> where is the place to put the defaults?

The places for reasonable defaults are the non-local settings files, i.e. `settings.yml`, `environment/#{envronment}.yml`. Those need to be overwritten (adjusted to the specific instance) in the `settings.local.yml` and similar.

> how can ranges for certain values be described (e.g. min max

Currently, that's not possible. Could you give me a pointer to settings that need it?

> how can be system properties access? E.g. to define a reasonable default like #{numcpus} * 0.75

Same question.

> for those few settings a simple check at a central place like already shown in the1170-settings-cleanup.patch would have saved you probably a lot of time

As stated in the pull request, that's not the place for validations or setting defaults. (according to rails). That's the place for reading the settings and configuring some rails-internals. 

> So, forking a different gem, writing about 300 lines of code, which doesn't provide, what is actually needed, introducing other errors/potential inconsistencies is for my taste a little bit too much.

There are no solutions for the settings-validation task out there, so we needed to create one ourselves. I know this is not the most elegant one, but it is a central place for validation and for validation only. I would have liked to use `ActiveModel::Validations` for the validations because it already defines some neat validation-methods. After playing around a little bit with this class, I figured out that it needs a lot more code than what this pull request contains.

The fork is not a that big a deal. The important changes are in a pull request of the original gem anyway. It just needs to be reviewed and merged. Then we can use the original rails_config again.

config/application.rb
eugenk: Have you tested it? On my machine the `settings.local.yml` has overwritten the preferences set in `hets.yml`.

lib/settings_validator.rb
jelmd: a) skip meta - doesn't change anything. As a starting point [wikipedia] (http://en.wikipedia.org/wiki/XML_schema) might be sufficient.

b) No, not at all! Again, reasonable defaults are a application internal thing! When a usual user can change it to whatever he wants, why should they still be considered reasonable. When you say *settings* you actually give away the control completely and can't guarantee anything! The only public place, where they should appear is in the documentation! BTW, you flipflop a lot - one time you say, overwrite by a 2nd file is "not clean", the next time you say, one should overwrite using a 2nd file. Looks like a big confusion to me.

c) You see, re-inventing the wheel doesn't make sense unless it is really worth the burden - in the industry they would ask you: show me your business case. If you than tell the manager, that you just need to check 1-2 dozen of app parameters, he'll say, what? Because you are to lazy to write down ~3 simple LOCs/param you wanna create another many man-month project, which probably no one will use, because writing down the description for the parameters takes at least the same time and probably not much less LOCs than when just doing it the usual, simple way? "Kanonenkugeln und Spatzen ..."

c,d) The point is: your class covers the very basic things, ~10% , 90% are still missing. So you made the application more complex and harder to understand but failed to actually solv the problem. An example for e.g. min max is hets workers.

e) Sorry, but reading and validating data read should be one step. The paradigm which goes hand in hand with this is "fail fast"! And please, keep focused on the task. I didn't say that's the right place do it, I said, that's what you could have been used (and implicit what experienced programmers would have been done) without wasting a lot of your and _my_ time! Also I get tiered about your "that's how rails tells one should do". I don't believe this non-sense anymore - so if you make such statements it would be nice to give a reference for this.

f) data validation is always a boring business, but needs to be done and wrt. to app parameters  it is damn easy. More or less a copy and paste thing - I say for the most cases a blind janitor could do that. So I could not understand at all, why you make a mountain out of a molehill ...

g) But enough to introduce other bugs ...

BTW: That's it for me. I supplied proper/workable suggestions and if you don't wanna take it, make what you want. If you want it to be stupid, keep it stupid. I'm scared of your NIH attitude.

eugenk:
I have taken the liberty to put the discussion here so we don't have three places to talk about the code.

##### As a follow-up on https://github.com/ontohub/ontohub/issues/1231#issuecomment-76883975:

1) I know what you meant with changing the data type. There are two possibilities for that:
* a) Preserving the datatype: We would use three settings files. The basic one (let's say settings.yml) sets an array, the second one (settings.production.yml) sets a string and then the third one (settings.local.yml) sets an array again. This is overly complicated for the one who writes the custom settings. And we can't expect him to know the internals of the settings gem.
* b) Actually changing the data type: We would use only two settings files. The basic one (settings.yml) sets an array and the next one (settings.local.yml) sets a string. This means we need to have code that can handle both. That code would be less readable and maintaining it wouldn't be a lot of fun.

This is why I want to introduce the behaviour that arrays get overwritten and not merged. It's basically what I would expect when I write a custom settings file.

And yes, I have studied the code and I know what deep_merge does. I missed the line in the rails_config.rb with the condition whether Rails is defined, but I spent much time finding out what the actual business logic of the gem does.


2) I think one can't really compare ontohub to an ordinary smartphone app. The users of ontohub (the ones who visit ontohub.org) don't need to configure anything. Who needs to configure something are the people who setup an instance. And as we said, the only files where actual configuration takes place, are the following:
```
config/settings.local.yml
config/settings/#{environment}.local.yml
config/environments/#{environment}.local.rb
```
And it perfectly makes sense to have three files for the task because sometimes you need to set/change something static only in a specific environment (e.g. production, development, test) and sometimes settings actually need computation (the rb file). The other files are supposed to set defaults and are only to be touched by the developers.

3) I have taken a look at your patches and as I said in the pull requests there, the idea of it is good, but we need to separate concerns (in this case setting, reading, validating and using the settings). Also, maintaining that code is not made easy because there are a lot of temporary variables with names that make the code hard to understand.

##### As a follow-up on https://github.com/ontohub/ontohub/pull/1236/files#r25663639:

a) Using XML, one can set a grammar and do some validation with that, yes. However, in Rails we use yaml. And yaml can't have a schema. The reason to use yaml is just simplicity.

b) As I said before, the following files are for defaults, set by the developers:
```
config/settings.yml
config/settings/#{environment}.yml
config/environments/#{environment}.rb
```
And the other (`*.local.*`) are to be used to write custom settings.
> one time you say, overwrite by a 2nd file is "not clean", the next time you say, one should overwrite using a 2nd file

I meant what I say in this post in 1). Sorry that I didn't express myself clearly.

c) If the wheel has weak spots (e.g. is oval and not round), there is good reason to reinvent it. This is supposed to be a tool to easily add more validations when more settings are introduced.

d) That's why we have a pull request and don't merge by ourselves. If code in my pull requests is really too complex or there is something else wrong with it, I'm gladly changing it to make it better.

e) "Fail fast" is a good principle, but there are several approaches to it. One can fail separately for each validated setting and present the user (admin in this case) only one error at a time, or one can gather all errors and present all of them before the application has fully booted. I chose the latter to make the admin's work easier. If you show me a pointer to where we need to fail before the initialization, I will change the behaviour.

lib/settings_validator.rb
eugenk: I replied on this in https://github.com/ontohub/ontohub/pull/1236#issuecomment-76915377

config/application.rb
eugenk: Oh and `reload!` is the only way to load another file. Simply adding it as a config file does not load it.

eugenk:
Now this should set reasonable defaults for the settings and validate the settings in the same extent as in  b8787b52f71a1f8922aff0c049a06c54ac60510c, but with everything in their distinct places.

@jelmd: In b8787b52f71a1f8922aff0c049a06c54ac60510c, there were some things that showed a misunderstanding:
* `Settings.external_repository_name` is not a fallback. It's a name of an ontohub-repository. I documented this in the file. See 2f4cb1ae60d4fcd0d196ae19fb66563a8f61dbf3.
* We actually want to send mails in the `test` environment with the `:test` delivery method. It doesn't *really* send a mail. It only allows to test if a mail would have been sent.

I'm pretty sure this is clear to you, but I want to be absolutely sure there's no misunderstanding: We want the `settings.yml`, `settings.<environment>.yml`, and `environments/<envoronment>.rb` to set reasonable default values. The configuration for the concrete ontohub instance (develop.ontohub.org, staging.ontohub.org, ontohub.org) is supposed to take place in the `*.local.*` files, which are not part of the feature-, develop, staging and master branches. They are supposed to be force-added to the branches develop.ontohub.org, staging.ontohub.org and ontohub.org.

ebolloff:
When I delete the `true`/`false` in e.g `production.rb`, so `config.assets.digest = ` instead of `config.assets.digest = true`, there's no validation error when I run `script/rebuild-ontohub`. Shouldn't there be one? Like 'Hey, you forgot the value'? Or is there a reason for it?

eugenk:
The production.rb does not get loaded in your development environment. If you put the same line into your development.rb, rails fails because of a syntax error (no value after the `=`). If you set `config.assets.digest = nil`, rails boots up just fine. However, you say, you want this to be validated as well. I will add some validations.

My first thought, when I wrote the validations, was to only validate what's specific to ontohub. Those are  Rails settings and exist in every Ralis app. If we validate anything, though, we should validate everything, including those Rails settings.

eugenk:
On second thought, it's not quite feasible to validate the Rails (and gem's) settings as well, because there's a whole lot of them and Rails (and the gems) can add some allowed values without us knowing about it. If there are some values, you explicitly want to be validated, I will gladly add validations. Otherwise I think it's somewhat of a validation overkill.

0robustus1:
We need to clearly separate what's internal and what's a public API (or better ACI, for Configuration).
If it is not public (and as such not meant to be changed by an admin/user) we should not do the work of validating it.

ebolloff:
When the to-validate and not-to-validate things are separated, I will go on reviewing this.

eugenk:
I documented the settings loading pipeline and the validations in the productive deployment doc file.

tillmo:
Is it correct that `/home/git/` no longer occurs in `settings.yml`?

eugenk:
Well in #1268 it appears and is needed for two reasons:
  1. It is read in when `rake git:compile_cp_keys` is executed and the value is being hardcoded into the `cp_keys` binary
  2. We need a value other than `~git` (namely `tmp/git`) in our development and test environments because we want those environments to be self-contained and not to mess with the whole system.

tillmo:
OK, that is in accordance with what @jelmd told me.

eugenk:
I just rebased this branch on staging.

ebolloff:
I found a few things, but you fixed them on other prs so :+1:

PR #1238, opened 2015-03-04
re-enqueue only specific container if not terminal

0robustus1:
If the ontology state is not terminal (a final state), then we should
only re-enqueue the particular container instead of all (which would
lead to a lot more jobs being created).


ebolloff:
:+1:

PR #1240, opened 2015-03-04
1166 add mimetype info to controllers

### LINKED ISSUE #1166: Add supported mime-types for every controller
We should declare supported mime-types for every controller/action in order to return a 406 response if a mime-type is not available.

0robustus1:
I've reduced that to the API relevant controllers, as i'm not entirely sure what all controllers should actually respond to. So this will close #1166.


eugenk:
:+1:

PR #1243, opened 2015-03-05
Fix ignoring local config files.

eugenk:
The settings files in config/environments/ are ruby files and not yml. Those locals need to be ignored.


ebolloff:
:+1:  when tests are green.

PR #1247, opened 2015-03-07
1233 json for model resources

0robustus1:
Shall close #1233. I did also throw in serialization for ontology type as an additional gimmick. Project and Task will probably follow in a later issue/PR.

Ah one additional note, this Pull Request is based on the **1166-add_mimetype_info_to_controllers** branch instead of **staging**. So one might want to review (and merge) #1240 before tackling this pr.

PR #1248, opened 2015-03-08
1246 use api controllers

### LINKED ISSUE #1246: Group API versions together with constraint-namespace
In order the be able to lock in implementations for specific versions
of the API, we'll need to maintain a separate Namespace of (at least)
controllers. There is some old code inside the application which was intended
to that. As we however won't be using route-namespacing or a subdomain for the API
we'll need to group the routing for the API by enforcing a constraint-namespace:
(these are conditions which will be `or`-ed together)

- API-Header is set
- Requested response is JSON

Additionally we'll need to extract existing API code into the namespace.

0robustus1:
Shall close #1246. My first idea (laid out in the issue-description) was not really appropriate, as we would've
needed to create multiple definitions (at least one for every api-version and one for the frontend) in order to provide outer grouping. Instead we'll do 'inner grouping' by extending the DSL in order to reroute
on mime-type (or arbitrary header, which is not in use yet) to another controller, in this case the api provided controller instead of the default frontend controller.

Additionally this moves every api-action into the appropriate api controller. This is based on the **1233-json_for_model_resources** branch (which is in turn based on **1166-add_mimetype_info_to_controllers**).
So please obey the following review-order:

1. #1240 
2. #1247 
3. #1248 (this one)


eugenk:
The seeds fail at the proof statuses step: Hets can't retrieve the proof statuses file from the running ontohub instance.

```bash
❯ wget --header='Accept: */*; q=0.1, text/plain' --no-check-certificate -O - http://localhost:3000/ref/1/meta/proof_statuses 2> /dev/null
{"iri":"localhost:3000/ref/1/meta/proof_statuses","evaluation_state":"failed","number":1,"commit_oid":"0ae3c897ea23e8729f681c147d6fd19006c0c879","basepath":"proof_statuses","file_extension":".owl","ontology":{"iri":"localhost:3000/meta/proof_statuses","name":"Proof_statuses"}}
```
(The wget call is copied from the Hets source)

eugenk:
http://ontohub.dev/meta should open the file browser, but it shows the repository show-page.

0robustus1:
Jep that is correct.

0robustus1:
http://ontohub.dev/meta should now again show the file browser.
Additionally the seed-problem should be fixed.

eugenk:
Yes, it's fixed. I'm reviewing the last commits right now.

eugenk:
I'm not sure if this is a problem with curl, specroutes, rails routing or if this is a request for whom no support is intended:
```
❯ curl -H 'Accept: application/json' http://ontohub.dev/default/partial_order//strict_partial_order//irreflexive
ActiveRecord::RecordNotFound at /default/partial_order/strict_partial_order/irreflexive
=======================================================================================

> Couldn't find Sentence without an ID
[...]
```
The same goes for http://ontohub.dev/default/partial_order//strict_partial_order//elem

0robustus1:
Jep, when i adjusted the routes for the MIME-type override i forgot to set the id in the router-constraint. So it is actually an 'issue' in rails routing ;)

eugenk:
:+1: after merging #1247 and rebasing staging

PR #1259, opened 2015-03-11
Delegate locid to Sentence

### LINKED ISSUE #1258: LocId missing on TranslatedSentence
http://staging.ontohub.org/bug-reports/double_import_blendoid//DIB-CarsAreAudis///sentences?all=true is failing with:
```
NoMethodError at /bug-reports/double_import_blendoid/DIB-CarsAreAudis/sentences
undefined method `locid' for #<TranslatedSentence:0x007f889f3a5b18>
```

eugenk:
This shall fix #1258.

### DIFF
--- a/app/models/translated_sentence.rb
+++ b/app/models/translated_sentence.rb
@@ -7,7 +7,7 @@ class TranslatedSentence < ActiveRecord::Base
   attr_accessible :audience, :ontology, :sentence, :symbol_mapping
   attr_accessible :translated_text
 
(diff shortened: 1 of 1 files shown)


0robustus1:
This would yield the locid of the base-sentence as the locid of the translated-sentence. So to the user it wouldn't really be accessible. However, let's accept this as a hotfix, but in order to actually close #1258 we'll need to create real loc/ids for translated sentences and probably create a custom serializer.
:+1:

PR #1260, opened 2015-03-11
1257 introduce axiom model

### LINKED ISSUE #1257: Introduce Axiom model
Currently, we have the models `Sentence` and `Theorem < Sentence`. To get a better distinction, we want to have `Axiom < Sentence`, which contains all Sentences that are not Theorems.

Update the views to show Axioms instead of Sentences.

Remove the default scope of Sentence.

eugenk:
This shall fix #1257. This pull request is based on the branch of #1188 (**1131-prove_theorems**) instead of **staging** . It might be easier to review that one first.


app/controllers/axioms_controller.rb
```diff
@@ -0,0 +1,41 @@
...
+        super
+      end
+  end
+
+  def logically_translated?
+    parent.contains_logic_translations?
+  end
+
+  helper_method :logically_translated?
+end
```
tillmo: acutally, axioms_controller.rb and sentences_controller.rb are identical, except that "axiom" is replaced by "sentence" at 5 places. Couldn't this be abstracted somehow?

app/models/ontology.rb
```diff
@@ -158,6 +158,12 @@ def all_sentences
       where('imported = ? OR imported = ?', true, false)
   end
 
+  def all_axioms
+    Axiom.unscoped.
+      where(ontology_id: self).
+      where('imported = ? OR imported = ?', true, false)
```
tillmo: what is the reason of this `where` clause, given that `imported boolean DEFAULT false NOT NULL`? Could `imported` be anything else than `true` or `false`?

app/controllers/axioms_controller.rb
eugenk: Follow-up question: Do we want to have a SentencesController for the web app (i.e. not the API) serving both, axioms and theorems? If not, I would like to remove it.

app/models/ontology.rb
eugenk: Fixed with ae691abf44f1d6477ddb8bf5fd7e3b32457f3b7a

app/serializers/axiom_serializer.rb
```diff
@@ -0,0 +1,23 @@
+class AxiomSerializer < ApplicationSerializer
+  class Reference < ApplicationSerializer
+    attributes :iri
+    attributes :name
+
+    def iri
+      qualified_locid_for(object)
+    end
+  end
```
tillmo: could the need for this class Reference (which occurs in all the serializers) please be explained somewhere?

spec/models/ontology_spec.rb
```diff
@@ -339,6 +339,26 @@
...
       end
     end
+
+    context 'axioms count' do
+      it 'should be correct' do
+        expect(ontology.axioms.count).to eq(1)
+      end
+
+      it 'should be reflected in the corresponding field' do
+        expect(ontology.axioms_count).to eq(ontology.axioms.count)
```
tillmo: why do you test something to be equal to itself? What does this test really test? non-nil?

app/serializers/axiom_serializer.rb
eugenk: When the API is called e.g. for an ontology, we don't want to include all the versions in the response. Instead, we send a reference (IRI) to an index of the versions. This index in turn contains only references to the versions themselves.

This does not only apply to versions, but also for all the other association ends our models have. In order to have a consistent way of describing and serializing those references, we have those Reference classes everywhere. See for example the line `has_one :ontology, serializer: OntologySerializer::Reference ` below.

spec/models/ontology_spec.rb
eugenk: This is not quite the same. On the left hand side, we have `axioms_count` (underscore) which is a column in the database and on the right hand side we have `axioms.count` (dot) which actually sends an sql query to count the rows.

app/controllers/axioms_controller.rb
tillmo: I think we don't need a SentencesController for serving sentences routes. We might need one if we want to abstract from both AxiomsController and TheoremsController. However, currently this seems not be necessary.

eugenk: Alright, then I'll remove the SentencesController, its routes and its views. I won't touch the API::SentencesController.

app/serializers/axiom_serializer.rb
tillmo: good. Could this explanation please be added in the code (or the doc) somewhere?

spec/models/ontology_spec.rb
tillmo: oh, sorry, I have overlooked this subtle difference.

tillmo:
Since I do not get notified about the commits, a short message would be helpful. 

Everything looks good, but one point is still open: "Could this explanation please be added in the code (or the doc) somewhere?"

eugenk:
> Could this explanation please be added in the code (or the doc) somewhere?

I have just completed it in #1288.

tillmo:
:+1:

PR #1261, opened 2015-03-12
1250 use hets api and vcr for the tests

### LINKED ISSUE #1250: Use Hets API and VCR for the tests
Currently, we use Hets to generate test fixtures for our ontology xml (later: json) files.  Use the Hets API to retrieve those xml files and record the responses with VCR. Use those recordings in the tests.

eugenk:
This shall fix #1250.

If no hets server is running, it is started and the fixture files (VCR cassettes) are created. Those fixtures are then used in the specs.

This pull request is based on the branch of #1260 (**1257-introduce_axiom_model**) instead of **staging**. It might be easier to review that one first.

### DIFF
--- a/lib/tasks/test.rake
+++ b/lib/tasks/test.rake
@@ -9,86 +9,177 @@ namespace :test do
     Rake::Task['db:migrate:clean'].invoke
   end
 
-  def hets_path
-    `which hets`.strip
-  end
 
-  def hets_out_file_for(ontology_file)
-    basename = File.basename(ontology_file).sub(/\.[^.]+$/, '.xml')
-    File.join('spec/fixtures/ontologies/hets-out/', basename)
-  end
+  HETS_API_OPTIONS = '/auto'
+  HETS_BASE_IRI = 'http://localhost:8000'
+  HETS_PATH = `which hets`.strip
+  HETS_SERVER_ARGS = YAML.load(File.open('config/hets.yml'))['server_options']
 
-  def on_outdated_files(files, &block)
-    files.each do |file|
-      hets_file = hets_out_file_for(file)
-      out_of_date = !FileUtils.uptodate?(hets_file, Array(hets_path))
-      block.call(file) if out_of_date
-    end
-  end
-
-  def hets_args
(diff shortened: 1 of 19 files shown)


eugenk:
This is now based on staging.

lib/tasks/test.rake
```diff
@@ -9,86 +9,197 @@ namespace :test do
...
+  def call_hets_provers(file, subdir)
+    puts "Calling hets/provers on #{file.inspect}"
+    options = 'format=json'
+    escaped_iri = Rack::Utils.escape_path("file://#{absolute_filepath(file)}")
+    hets_iri = "#{HETS_BASE_IRI}/provers/#{escaped_iri}#{HETS_API_OPTIONS}"
+    hets_iri << "?#{options}"
+
+    FileUtils.rm_f(recorded_file(subdir, file))
+    VCR.use_cassette(cassette_path_in_fixtures(subdir, file)) do
+      Net::HTTP.get_response(URI(hets_iri))
```
tillmo: maybe part of the code can be abstracted here, since the call_hets_* methods are quite similar. Cf. also lib/hets/*_caller.rb

lib/tasks/test.rake
```diff
@@ -9,86 +9,197 @@ namespace :test do
...
+      end
+    end
+  rescue Timeout::Error
+    false
+  end
 
-    filename = File.basename(file).split('.')[0..-2].join('.')
-    target_path = Rails.root.join('spec/fixtures/ontologies/hets-out/prove', filename)
+  def with_running_hets(&block)
+    hets_was_already_running = !port_open?('127.0.0.1', 8000)
```
tillmo: this seems to check whether Hets is *not* running

eugenk: Right, thanks. This is a bad variable name. I fixed it in f46f2125b8cdf6d8240c1b85b438f1fbc51c2da8.

lib/tasks/test.rake
eugenk: Done with c4be757e81d9c807502df9f21966fcbbde7c688e

tillmo:
:+1: modulo travis

PR #1268, opened 2015-03-17
1170 git handling

eugenk:
This pull request replaces #1197. What it does:
* Make the paths around git configurable.
* Add rake tasks to prepare git authorized_keys handling
  * Compile the key-copy binary (and set custom permissions)
    * Before compilation, the source file `cp_keys.c` is changed to use the settings which the Rails app uses. Make sure, this rake task is called with the correct RAILS_ENV.
  * Create the ontohub-editable authorized_keys with custom permissions file if it doesn't exist
  * Print messages on what is yet to do manually (`chown`). We can't do this automatically, because we donÄt know which owner/group we want to set. We could add settings for user/group, but we would to have need root privileges on some systems for `chown`. The root privileges are the reason why I want to let the admin do this manually.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -81,7 +81,7 @@ paths:
   commits: data/commits
   # home directory of the git user
   #   needed for handling of ssh keys in ~git/.ssh/authorized_keys
-  git_home: /home/git
+  git_home: ~git
 
 git:
   verify_url: http://localhost/
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -79,6 +79,9 @@ paths:
   symlinks: data/git_daemon
   # cache for files that needed to be checked out from the git repositories
   commits: data/commits
+  # home directory of the git user
+  #   needed for handling of ssh keys in ~git/.ssh/authorized_keys
+  git_home: /home/git
 
 git:
   user: # can be nil if git commands are to be executed with the current user
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -84,8 +84,6 @@ paths:
   git_home: /home/git
 
 git:
-  user: # can be nil if git commands are to be executed with the current user
-  group: # can be nil if git commands are to be executed with the current group
   verify_url: http://localhost/
   default_branch: 'master'
   push_priority:
--- a/.gitignore
+++ b/.gitignore
@@ -64,3 +64,13 @@ config/environments/*.local.rb
 /coverage
 
 spec/fixtures/vcr/hets-out
+# Ignore custom compiled binaries
+bin/cp_keys
+
+spec/fixtures/ontologies/xml
+
+spec/fixtures/ontologies/hets-out/*
+!spec/fixtures/ontologies/hets-out/.gitkeep
+
+spec/fixtures/ontologies/hets-out/prove
+!spec/fixtures/ontologies/hets-out/prove/.gitkeep
--- a/config/initializers/paths.rb
+++ b/config/initializers/paths.rb
@@ -19,17 +19,6 @@ module PathsInitializer
 
       config.commits_path =
         cleanup_release(Rails.root.join(Settings.paths.commits))
-
-      settings = Settings.git
-      if settings && settings.user
-        config.git_user  = settings.user
-        config.git_group = settings.group
-        config.git_home  = File.expand_path("~#{config.git_user}")
-      else
-        config.git_user  = nil
-        config.git_group = nil
-        config.git_home  = Rails.root.join('tmp','git')
-      end
     end
   end
 end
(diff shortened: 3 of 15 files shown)


jelmd:
:-1: As already said, this is just plain wrong.

eugenk:
Please elaborate

eugenk:
Oh, and when elaborating why this "is just plain wrong", please keep in mind that this pull request is only about the commits starting with the one having the message "Add cp_keys.c source file". The commits before (from #1236) are just needed here and this pull request won't be merged before #1236 gets merged.

jelmd:
"as said" usually means: already written in my previously submitted comments.
If you really wanna know what's wrong,  try to read and _understand_ the patches I made. And as said again, don't use any outdated stuff, but the recent ones made and already adjusted wrt. staging and actually _used_ on td. For your convenience  I've even mapped the related stuff (~ontohub/etc/) to http://td.iws.cs.ovgu.de/server-etc/ . If you have problems to understand it,  checkout e.g. to a11236c606909b2eb1d09e42aab0065a114268cb, apply all *.patch and try it out/debug it.

With these patches applied the only thing I've to do is to copy http://td.iws.cs.ovgu.de/server-etc/settings.local.yml into ~ontohub/onthub/config/ and build, i.e. ontohub works out of the box
(actually it would even work without any settings.local.yml correctly!!!).
If I take the current staging* w/o applying any patches, ontohub is broken - doesn't work.

And as said again, http://td.iws.cs.ovgu.de/server-etc/1170-settings-cleanup.patch shows, what I understand as "keep it small and simple". This is absolutely sufficient, easy to maintain/remember and does, what is needed!

eugenk:
tl;dr: Basically what I want say is: We need those patches be as small as possible (better: gone completely) and have them integrated into the ontohub code base in a (for rails developers) maintainable and easy to adjust way.


> don't use any outdated stuff, but the recent ones made and already adjusted wrt. staging and actually used on td

So your pull requests were already outdated, when I created #1236 and you kepts the patch files up to date? Then I need to revisit the current patches.

> With these patches applied the only thing I've to do is [...]

That's exactly the problem. As far as I understood Till (and totally agree), we need ontohub to work without any patches. Configuration (changing some yml files) is okay, but not patching application code. 

> And as said again, http://td.iws.cs.ovgu.de/server-etc/1170-settings-cleanup.patch shows, what I understand as "keep it small and simple". This is absolutely sufficient, easy to maintain/remember and does, what is needed!

I don't want to be rude or anything and please don't take it that way, but from my point of view this is not maintainable at all. Well, maybe you are able to maintain it, but I don't think that someone else can simply read the code and easily adjust it, if something needs to be changed some day. You say, our way is "over engineering", but the reason for it is to have everything in their distinct places with structure and to have (variable/method) names that mean something such that someone else can easily understand and adjust that code. We won't be ontohub-developers forever and when someone new comes, they should have the chance to understand *everything*. And coming back to you being able to maintain it: Maybe some day you are not available (e.g. illness, vacation, having another work place, etc.) and those patches need some adjustment. That person would probably need quite some time to do it.

tillmo:
note that this discussion is really about #1236 (the code of #1236 appears here, because #1268 is based on #1236).

config/settings.yml
```diff
@@ -54,7 +65,22 @@ exception_notifier:
...
+  data: data
+  # git repositories (names are numbers/ids)
+  git_repositories: data/repositories
+  # named symlinks to the git repositories
+  symlinks: data/git_daemon
+  # cache for files that needed to be checked out from the git repositories
+  commits: data/commits
+  # home directory of the git user
+  #   needed for handling of ssh keys in ~git/.ssh/authorized_keys
+  git_home: /home/git
```
eugenk: remove the setting `git_home` and use the expanded path `~git`.

tillmo:
@jelmd: this pull request introduces a settings variable `git_repositories`, as you requested. The `authorized_keys` file is always located in the home directory of the git user; this is standard in many projects. The `host_name` should not be automatically detected, since it is very important to the user of ontohub instances, and it could be set via a CNAME.

config/settings.yml
eugenk: Actually we at least need the settings key `git_home`. We want to set it to `tmp/git` in the development and test environment. The default value is now `~git` and it is not expanded.

eugenk:
I still need to do:
Document the rake task `git:compile_cp_keys` in the productive deployment.

eugenk:

ebolloff:
When #1330 is merged, please rebase this branch.

eugenk:
Done.

ebolloff:
Tests are red because of #1327, but green on my machine so :+1:

PR #1269, opened 2015-03-17
1170 add environment light with hets

### LINKED ISSUE #1272: Remove hets cli handling from hets.rb
Currently, we still have code for handling the hets command line in hets.rb, but we don't use it. Remove this dead code.

eugenk:
This pull request makes sure that the same Hets configuration is used in every place in which we need Hets.

Because we use a rails_config build with array-overwriting, we can overwrite the settings' `hets_path` in a later-loaded yml file and specify one specific binary (this is needed for security reasons).

##### Edit
Now this pull request also restructures the hets settings and removes hets cli handling (shall fix #1272).


0robustus1:
To be honest i do not really like the solution of just adding *hets.yml* to the settings-files.
There are a few reasons for that:

- hets settings should be in their own namespace inside of `Settings`, e.g. `Settings.hets.hets_path`, or `Settings.hets.hets_lib`.
- Additionally Settings like `version_minimum_version` or `version_minimum_revision` shouldn't be overrideable by a user/admin of an ontohub application
- The same applies to command-line-options and server-options. Both are only expected to be set by developers.

On a side-note, methods like `first_which_exists` can't handle `hets_path` being not an array right now.

I would suggest to move settings that should be configurable into the normal settings-files, into a *hets* namespace. This pretty much only includes the paths to hets, hets-lib, and hets-owl-tools and maybe the env-namespace. As a fallback (if neither of those settings is set) we should fallback to our search-paths for hets, hets-lib and hets-owl-tools.

eugenk:
You're right. Some settings are to be made by us only. I changed it.

> On a side-note, methods like first_which_exists can't handle hets_path being not an array right now.

There is a validation that expects it to be an array with at least one path to an executable. You can't supply anything but an array. Otherwise Rails won't boot.

eugenk:

eugenk:
I rebased this on staging. Tests on travis and jenkins are going to be red because of #1327.

0robustus1:
By my estimate: :+1:

ebolloff:
:+1:

PR #1275, opened 2015-03-19
Fix ontology state update

0robustus1:
Shall fix the ontology state feature yet again. This time for usage of locids, and in turn fixes version-locid resolving by excluding the query-string.

### DIFF
--- a/app/assets/javascripts/ontology_version-state.js.coffee
+++ b/app/assets/javascripts/ontology_version-state.js.coffee
@@ -8,7 +8,7 @@ update = (container) ->
   updateContainer = (container) ->
     current_state = container.data('state')
 
-    $.getJSON container.data('uri') + ".json", (data) ->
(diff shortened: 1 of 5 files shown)


eugenk:
I took the liberty to add some commits. Your code makes the right ajax calls, but there were a few things left to do:
1. The state wasn't read properly from the JSON response (`state` was always `undefined`).
2. The cucumber steps didn't make a check. Just for fun I replaced `find('span').has_content?(@state)` by `find('span').has_content?('foo')` and the test was still green. So I replaced it by an `expect` call.
3. Capybara can't know when to check for the changed content, because it only knows it when it can't find an element. The element is already existent and found, though, before the ajax response arrives. So we need to manually wait for the ajax to complete.

I give you :+1: if you give me :+1: ;)

0robustus1:
Sadly this seems to be the way to go when using poltergeist, so: :+1:

PR #1286, opened 2015-03-25
1273 improve proving interface

eugenk:
This shall fix #1273, except for the index page. The index page will part of #1283.

This branch is based on the branch of #1263 (**1237-let_the_user_select_a_prover**) instead of **staging**.  It might be easier to review that one first.

PR #1289, opened 2015-03-25
1256 add proving related serializers

### LINKED ISSUE #1256: Add sensible serializer for theorem
In #1188, a basic TheoremSerializer is added. It has the same functionality as SentenceSerializer. Improve it by adding information on the proof attempts etc.

##### New tasks (edit)
* [x] Add serializers listed below.
* [x] Add API controllers to serve those contents

### LINKED ISSUE #1290: Add specs on proving related API controllers
#1289 adds API controllers, but has no specs. Add specs for them.

eugenk:
This shall fix #1256, but it lacks specs. As soon as @0robustus1 specifies json schemata for those commands, I will adjust the serializers and API controllers (if needed) and add tests. This is another issue, though (#1290).

Because this also adds specs, this shall fix #1290 as well.

This branch is based on the branch of #1286 (**1273-improve_proving_interface**) instead of **staging**.  It might be easier to review that one first.


eugenk:
What's left to do after talking with @0robustus1:
* [ ] Add links from each ProofAttemptConfiguration to the using ProofAttempts.
* [ ] Move the prover output into a separate model.
* [ ] Write specs.

eugenk:
I stalled this pull request because I need to wait for the json schemata for logic mappings.

eugenk:
Now this pull request is ready for review. However, there is something that needs to be discussed:

On all three machines, we got
```ruby
LogicMapping.all.map(&:standardization_status).uniq
=> ["Unofficial"]
LogicMapping.all.map(&:defined_by).uniq
=> [nil]
LogicMapping.all.map(&:faithfulness).uniq
=> ["not_faithful"]
LogicMapping.all.map(&:projection).uniq
=> [nil]
LogicMapping.all.map(&:exactness).uniq
=> ["not_exact"]
LogicMapping.all.map(&:theoroidalness).uniq
=> [nil]
```
which are weird values - at least for `projection` (should be boolean) and for `theorodialness`. Maybe we should add a new issue to set reasonable values.

To be compliant to the json schema, I added a data migration to set `theorodialness` to `"unknown"`.


The to-do-item "Add links from each ProofAttemptConfiguration to the using ProofAttempts." from a previous comment is dropped. It's not part of the json schema. @0robustus1: Is this correct or do we want to reference the proof attempts from the proof attempt configuration again?

eugenk:
I just rebased on staging and amended 86adaab to work well with nil values.

eugenk:
I rebased this branch on staging. If you go through it commit by commit, you should know that 
* "Add locid to LogicMapping." 9b7cd92
* part of "Add API LogicMappingsController and its routes." 0583637

are reverted by the latest commits.

eugenk:
I just rebased staging and removed the code duplication of the slugs.

tillmo:
:+1: if travis goes through

eugenk:
Travis won't be green before #1327 gets merged ;)

tillmo:
I see, then :+1:

Issue #1156, opened 2015-01-12
Add creation of combined ontology to API


0robustus1:
Like #150 will create a user-interfaced version of combined ontology creation, the API should also
include a version of that.

0robustus1:
@tillmo This basically means creating a distributed ontology in DOL which will contain the combine-command, correct?
Where can i find the version of the DOL standard that is currently supported by hets?

tillmo:
correct. The DOL standard is available at https://github.com/tillmo/DOL

PR #1293, opened 2015-03-29
1292 add api key

### LINKED ISSUE #1292: Add API-Key
Users should be able to create (and use) API-keys.

0robustus1:
Shall close #1292.

Adds API-Keys to the application. Every user can create API-Keys. Only one per user will be valid at any given time. Currently the invalid ones will also be stored indefinitely - we might want to change this later on.

Also adds 'current_user'  handling based on the api-key for the API-Controller. Tests for this part will however have to wait until the first feature is implemented, that actually needs authentication (which will probably be #1156).

### DIFF
--- a/app/models/api_key.rb
+++ b/app/models/api_key.rb
@@ -1,0 +1,28 @@
+class ApiKey < ActiveRecord::Base
+  KEY_LENGTH = 48
+  STATES = %w(valid invalid)
+
+  belongs_to :user
+  attr_accessible :key, :status
+
+  before_validation :initialize_key
+
+  validates :user, presence: true
+  validates :key, presence: true, uniqueness: true
+  validates :status, inclusion: {in: STATES}
+
+  private
+  def initialize_key
+    generate_key! unless key
+    set_status! unless status
+    true
+  end
+
+  def generate_key!
+    self.key = SecureRandom.hex(KEY_LENGTH)
+  end
+
+  def set_status!
+    self.status = 'valid'
+  end
+end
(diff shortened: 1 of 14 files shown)


eugenk:
> Only one per user will be valid at any given time.

Why? One could want to have different API keys for different applications.

eugenk:
Jenkins, test this please. (I'm removing this comment as soon as the test job has started)

0robustus1:
API-Keys are user-based not application-based. Technically there is no reason to not allow multiple API-Keys, but this assumes that users can handle their keys correctly.
Later on Ontohub will probably also support OAuth (which is actually application-based).

eugenk:
The code looks good. I still need to play around with the new feature. Now I don't have time, though. Maybe later today or tomorrow in the morning.

eugenk:
Jenkins, test this please. (I will remove this comment as soon as the test job has started)

app/models/api_key.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def invalidate!
+    self.state = 'invalid'
+    save!
+  end
+
+  private
+  def initialize_key
+    generate_key! unless key
+    set_status! unless status
```
eugenk: Here we still have `status`

spec/factories/api_key_factory.rb
```diff
@@ -0,0 +1,9 @@
+FactoryGirl.define do
+  factory :api_key do
+    association :user
+
+    factory :invalid_api_key do
+      status { 'invalid' }
```
eugenk: Here we still have `status`

eugenk:
:+1:

PR #1297, opened 2015-03-31
1295 refactor proof evaluation pipeline

### LINKED ISSUE #1295: Refactor proof evaluation pipeline
We have a confusing class hierarchy with `Hets::Prove::ProveEvaluator` and `Hets::Prove::Evaluator` and the latter defines empty methods. This has been implemented while being pressed for time, but now this really needs to be refactored for better understanding.

eugenk:
This shall fix #1295. Other than the previous pull requests, this one is actually based on staging.


tillmo:
:+1:

PR #1298, opened 2015-04-01
1296 let the user select a proving timeout

### LINKED ISSUE #1296: Let the user select a proving timeout
The attribute already exists in `ProofAttemptConfiguration`. The timeout is, however, not selectable by the ontohub user.

eugenk:
This shall fix #1296.

### DIFF
--- a/spec/lib/proof_spec.rb
+++ b/spec/lib/proof_spec.rb
@@ -37,8 +37,8 @@ describe Proof do
     end
 
     it 'is not valid with bad provers' do
-      bad_params = params.merge({proof: {prover_ids: ['-1', '']}})
-      expect(Proof.new(bad_params).valid?).to be(false)
+      params[:proof][:prover_ids] = ['-1', '']
+      expect(Proof.new(params).valid?).to be(false)
     end
 
     it 'is not valid with a too little timeout' do
--- a/spec/lib/proof_spec.rb
+++ b/spec/lib/proof_spec.rb
@@ -146,6 +146,28 @@ describe Proof do
           end
         end
 
+        context 'have the provers set' do
+          # On OntologyVersion, there are created "the same" ProofAttempts for
+          # each contained Theorem. We cannot check for equality of the array,
+          # because there are duplicate entries (wtr. to provers).
+          it 'inclusion of pa-conf provers in defined provers' do
+            proof.proof_attempts.each do |proof_attempt|
+              expect(provers).
+                to include(proof_attempt.reload.
+                  proof_attempt_configuration.prover)
+            end
+          end
+
+          it 'inclusion of defined provers in pa-conf provers' do
+            pa_configuration_provers = proof.proof_attempts.map(&:reload).
+              map(&:proof_attempt_configuration).
+              map(&:prover)
+            provers.each do |prover|
+              expect(pa_configuration_provers).to include(prover)
+            end
+          end
+        end
+
         it 'have the timeout set' do
           proof.proof_attempts.each do |proof_attempt|
             expect(proof_attempt.reload.proof_attempt_configuration.timeout).
@@ -277,6 +299,15 @@ describe Proof do
             expect(proof_attempt.reload.proof_attempt_configuration).
               not_to be_nil
           end
+        end
+
+        it 'have the provers set' do
+          # For a single theorem we can check for equality because there are no
+          # more proof attempts with the same provers.
+          pa_configuration_provers = proof.proof_attempts.map(&:reload).
+            map(&:proof_attempt_configuration).
+            map(&:prover)
+          expect(pa_configuration_provers).to eq(provers)
         end
 
         it 'have the timeout set' do
--- a/spec/lib/proof_spec.rb
+++ b/spec/lib/proof_spec.rb
@@ -16,6 +16,8 @@ describe Proof do
               timeout: timeout}
     }
   end
+
+  let(:options) { { timeout: timeout} }
   before do
     ontology.theorems << theorem2
     ontology.save!
@@ -72,7 +74,10 @@ describe Proof do
 
     it 'prove_options_list contains the provers' do
       expect(proof.prove_options_list).
-        to eq(provers.map { |prover| Hets::ProveOptions.new(prover: prover) })
+        to eq(provers.map do |prover|
+          options[:prover] = prover
+          Hets::ProveOptions.new(options)
+        end)
     end
 
     it 'map ProveOptions to a list of "theorems count" many ProofAttempt ids' do
@@ -219,7 +224,10 @@ describe Proof do
 
     it 'prove_options_list contains the provers' do
       expect(proof.prove_options_list).
-        to eq(provers.map { |prover| Hets::ProveOptions.new(prover: prover) })
+        to eq(provers.map do |prover|
+          options[:prover] = prover
+          Hets::ProveOptions.new(options)
+        end)
     end
 
     it 'map ProveOptions to a list of "theorems count" many ProofAttempt ids' do
--- a/spec/lib/proof_spec.rb
+++ b/spec/lib/proof_spec.rb
@@ -2,6 +2,7 @@ require 'spec_helper'
 
 describe Proof do
   let(:provers) { [1,2].map { create(:prover, :with_sequenced_name) } }
+  let(:timeout) { 10 }
   let(:theorem) { create :theorem }
   let(:ontology) { theorem.ontology }
   let(:theorem2) { create :theorem, ontology: ontology }
@@ -11,7 +12,8 @@ describe Proof do
     {
       repository_id: repository.to_param,
       ontology_id: ontology.to_param,
-      proof: {prover_ids: [*provers.map(&:id).map(&:to_s), '']},
+      proof: {prover_ids: [*provers.map(&:id).map(&:to_s), ''],
+              timeout: timeout}
     }
   end
   before do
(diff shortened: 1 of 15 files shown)


tillmo:
The timeout works fine. However, please change the timeout selection box. First, it is too wide - it should be only (roughly) as wide as the text it contains. Second, selecting a timeout of 1 sec does not make sense. I propose the following timeouts: 10s, 30s, 1min, 5min, 10min, 30min, 1h, 6h, 1d, 2d, 7d. If the server load becomes to high, we might restrict the longer timeouts, but I suggest to start with these.

spec/lib/proof_spec.rb
```diff
@@ -229,6 +299,33 @@
...
 
+      context 'ProofAttemptConfigurations' do
+        before { proof.save! }
+
+        it 'are created' do
+          proof.proof_attempts.each do |proof_attempt|
+            expect(proof_attempt.reload.proof_attempt_configuration).
+              not_to be_nil
+          end
+        end
```
tillmo: this is exactly the same code as in line 144ff.

spec/lib/proof_spec.rb
```diff
@@ -229,6 +299,33 @@
...
+
+        it 'have the provers set' do
+          # For a single theorem we can check for equality because there are no
+          # more proof attempts with the same provers.
+          pa_configuration_provers = proof.proof_attempts.map(&:reload).
+            map(&:proof_attempt_configuration).
+            map(&:prover)
+          expect(pa_configuration_provers).to eq(provers)
+        end
```
tillmo: this is code similar to that in line 154ff. Please check.

spec/lib/proof_spec.rb
```diff
@@ -229,6 +299,33 @@
...
+
+        it 'have the timeout set' do
+          proof.proof_attempts.each do |proof_attempt|
+            expect(proof_attempt.reload.proof_attempt_configuration.timeout).
+              to eq(timeout)
+          end
+        end
+      end
+
```
tillmo: this is exactly the same code as in line 176ff.

eugenk:
I'm not sure if the (really) long timeouts are going to help at all. In all the ATP-related papers I read, they were using timeouts of a minute (at most). There also was a paper stating that more than a few seconds (I don't know the value anymore but I guess this was around 10 seconds) doesn't allow for more goals to be proven. I can add the longer timeouts anyway if you want to have them.

As for the test code:
Yes, this is the same code, but with different premises. The function is called once on `ontology_version` and once on `theorem`. For some examples the outcome is supposed to be the same and for some examples it is supposed to be similar. I could have created a separate context where the outcome is the same, but I wanted to make the test code more readable because this is also sort of a specification.

tillmo:
With the Vickrey auction theorem, I indeed had the experience that ATP sometimes find more with larger timeouts. Maybe your axiom selection algorithm will make larger timeouts superfluous, but until then, I would like to have them. The sledgehammer paper says that SPASS will find any proof either in the first 10 sec or not at all, but I am not sure whether this holds for all provers.

Concerning the code duplication, I follow your decision.

eugenk:
7068bf0 adds longer timeouts.

tillmo:
7068bf0 looks good, but still the selection box is too wide

eugenk:
756dcaa should make the select box smaller.

tillmo:
Thanks for the smaller box.

Problem: longer timeouts do not work. Take http://develop.ontohub.org/sandbox/Vickrey_proof.het. On my local machine, I try to prove vickreyA_pre01 with SPASS and 5min timeout. Indeed, then SPASS runs for a couple of minutes. However, after SPASS has terminated, the prover output is not stored in the database.

eugenk:
Right, we get a `Net::ReadTimeout` in our POST request to Hets. I can set this timeout higher as a quick fix, but the real fix would be to change the behaviour of Hets to:

1. A prove request is sent with POST
2. Hets responds with HTTP 202 (Accepted), an ETA and a URL to where the one can check the current status (processing/done)
3. Ontohub checks back after some time at this URL
  1. If the current status is still processing, repeat step 3.
  2. If the current status is done, Hets responds with a URL to the final resource (the JSON data we already get from Hets)
4. Ontohub processes the JSON data.

tillmo:
OK, please set the timeout higher and open a Hets issue, then :+1:

eugenk:
Now a http timeout is being set (depending on the timeout of the ontohub proof command). Right now I have no time to test it with higher timeouts set in the proof command. I will try this today in the evening.

eugenk:
And I filed a Hets issue at https://github.com/spechub/Hets/issues/1481

tillmo:
for me, a 5 min timeout still did not work.

eugenk:
There seems to be a bug of Hets involved. **We actually get a response after exactly one minute**, but this is a `500 Internal Server Error` with the content `Something went wrong`. Hets does not output anything on the console.

This also happens if there is no rails server involved **and a timeout is set to two minutes**:
```
❯ curl -v --max-time 120 -X POST -H "Content-Type: application/json" -d '{"format":"json","include":"true","prover":"SPASS","timeout":"300","node":"sp"}' http://localhost:8000/prove/file%3A%2F%2F%2Ftmp%2FVickrey_proof.het/auto
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 8000 (#0)
> POST /prove/file%3A%2F%2F%2Ftmp%2FVickrey_proof.het/auto HTTP/1.1
> User-Agent: curl/7.37.1
> Host: localhost:8000
> Accept: */*
> Content-Type: application/json
> Content-Length: 79
>
* upload completely sent off: 79 out of 79 bytes
* HTTP 1.0, assume close after body
< HTTP/1.0 500 Internal Server Error
< Date: Thu, 02 Apr 2015 19:37:23 GMT
< Server: Warp/3.0.1.1
< Content-Type: text/plain; charset=utf-8
<
* Closing connection 0
Something went wrong%
>>> elapsed time 1m0s
```

What happens in Ontohub:
Sidekiq reports a `stack level too deep` after an error is thrown:
```
UriFetcher::Errors::UnfollowableResponseError: can't follow the response, and response not useable
from ontohub/lib/uri_fetcher/http_caller.rb:67:in `block in handle_response'
```

eugenk:
I filed https://github.com/spechub/Hets/issues/1482 for the first part. For the second part, I will change ontohub to handle the `stack level too deep` error.

eugenk:
Sending the proof command synchronously (i.e. calling `CollectiveProofAttemptWorker.new.perform(...)` in the rails console), everything works just fine. Using an asynchronous call with sidekiq, we get a `stack level too deep` error.

Using [this method to create a trace](https://gist.github.com/jbgo/4493822) (12 MB of text), I found that sidekiq loops in the method `Sidekiq::Middleware::Server::RetryJobs#exception_caused_by_shutdown?` ([see the code](https://github.com/mperham/sidekiq/blob/master/lib/sidekiq/middleware/server/retry_jobs.rb#L195-196)). This code descends in the error cause hierarchy until there is either no cause or a sidekiq shutdown as cause.

Then I checked our code with the synchronous call again, catching the the error:
```
[1] pry(main)> CollectiveProofAttemptWorker.new.perform(...)
<breakpoint in a rescue block>
[1] pry(#<CollectiveProofAttempt>)> e
=> #<UriFetcher::Errors::UnfollowableResponseError: can't follow the response, and response not useable>
[2] pry(#<CollectiveProofAttempt>)> e.cause
=> #<Hets::Errors::NotAHetsError: Error was not produced due to hets.>
[3] pry(#<CollectiveProofAttempt>)> e.cause.cause
=> #<UriFetcher::Errors::UnfollowableResponseError: can't follow the response, and response not useable>
[4] pry(#<CollectiveProofAttempt>)> e.object_id == e.cause.cause.object_id
=> true
```
So we have a cycle in the causes. This leads to an infinite recursion in sidekiq.

eugenk:
For now, I try to remove the circular causes in our code (which will be more of a hack than a clean solution). I also filed https://github.com/mperham/sidekiq/issues/2284. If this gets handled in sidekiq, I will remove the hack in our code again.

eugenk:
I just tried this locally. Now the correct error (can't follow the response) is displayed in the Sidekiq UI and in the ontology versions view (e.g. http://ontohub.dev/default/Vickrey_proof//sp///ontology_versions).

Since Hets closes the connection after a minute, I suppose I shall remove the timeouts longer than 30 seconds. Is this correct? Maybe add one for 45 seconds.

Also, Hets *always* closes the connection after a minute, even if there are many goals and it has stopped working on the first goal after 45 seconds and started working on the next one. Hets also considers the timeout parameter as a timeout for each goal. So I think we should consider the selected timeout as a total timeout in Ontohub and pass `timeout/4` to Hets if there are four goals.

tillmo:
agreed to divide the timeout by the number of goals. Maybe you can display a mouseover (over "Timeout") explaining that the time will be used for all goals altogether.

Aha, and I would prefer the default timeout and the default prover, e.g. "by default, the timeout is 10s" instead of "Selecting none will result in using the default timeout."

eugenk:
Is SPASS the overall default prover or are there cases in which another prover is set by default?

tillmo:
For higher-order logic (HasCASL, THF) SPASS wouldn't work. The default prover depends on the logic. So probably Hets should mark the default prover in the list of provers.

eugenk:
I created #1301 for the provers. The timeout hint is changed in 11d1e7d.

tillmo:
:+1:

PR #1300, opened 2015-04-02
Put fallbacks into user_info, use user_info.

### LINKED ISSUE #1282: Repository::Git doesn't use user_info consequently
The methods `delete_file` and `save_file` use `user_info(user)` while `save_file_only` doesn't. The latter is the only method that uses fallbacks if no user is supplied. This needs to be made consistent.

eugenk:
Shall fix #1282. This is just a small refactoring.


0robustus1:
:+1:

PR #1308, opened 2015-04-08
Abort test execution unless elasticsearch is running.

### LINKED ISSUE #1303: Let default rake task fail if elasticsearch is not running
Elasticsearch running is a dependency of the default rake task. If it is not running, some tests are red. Let the task fail immediately if this dependency is not met.

eugenk:
This shall fix #1303.

### DIFF
--- a/lib/tasks/test.rake
+++ b/lib/tasks/test.rake
@@ -167,6 +167,17 @@ namespace :test do
     end
   end
 
+  desc 'abort execution if elasticsearch is not running'
+  task :abort_if_elasticsearch_is_not_running do
+    elasticsearch_port = ENV['ELASTIC_TEST_PORT'].present?
+    elasticsearch_port = '9250' unless elasticsearch_port.present?
+    unless port_open?('127.0.0.1', elasticsearch_port)
+      message =
+        'Elasticsearch is not running. Please start it before running the tests'
+      raise message
+    end
+  end
+
   desc 'Update all fixtures'
   task :freshen_fixtures do
     outdated_exist = outdated_cassettes(ontology_files, 'dg').any?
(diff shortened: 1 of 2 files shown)


lib/tasks/test.rake
```diff
@@ -167,6 +167,17 @@ namespace :test do
     end
   end
 
+  desc 'abort execution if elasticsearch is not running'
+  task :abort_if_elasticsearch_is_not_running do
+    elasticsearch_port = ENV['ELASTIC_TEST_PORT'].present?
+    elasticsearch_port = '9250' unless elasticsearch_port.present?
+    unless port_open?('127.0.0.1', elasticsearch_port)
```
0robustus1: if `ENV['ELASTIC_TEST_PORT']` is actually `present?` the value of `elasticsearch_port` will be `true`, which is probably not what you want.

eugenk: Thanks, I was too quick there. Should be fixed by fa1ef27.

lib/tasks/test.rake
```diff
@@ -167,6 +167,17 @@ namespace :test do
...
   end
 
+  desc 'abort execution if elasticsearch is not running'
+  task :abort_if_elasticsearch_is_not_running do
+    elasticsearch_port = ENV['ELASTIC_TEST_PORT']
+    elasticsearch_port = '9250' unless elasticsearch_port.present?
+    unless port_open?('127.0.0.1', elasticsearch_port)
+      message =
+        'Elasticsearch is not running. Please start it before running the tests'
+      raise message
```
0robustus1: Why do you raise an error instead of `puts`-ing to STDERR and `exit 1`?

eugenk: good point

0robustus1:
:+1:

PR #1309, opened 2015-04-08
Use OS hostname if hostname is undefined in settings

eugenk:
This automatically reads the hostname from the operating system and uses it in the application if `hostname` is not set in the settings.yml (as @jelmd requested).

### DIFF
--- a/app/serializers/application_serializer.rb
+++ b/app/serializers/application_serializer.rb
@@ -4,7 +4,7 @@ class ApplicationSerializer < ActiveModel::Serializer
   end
 
   def qualified_locid_for(resource, *commands, **query_components)
-    iri = "#{Settings.hostname}#{resource.locid}"
+    iri = "#{Ontohub::Application.config.fqdn}#{resource.locid}"
     iri << "///#{commands.join('///')}" if commands.any?
     iri << "?#{query_components.to_query}" if query_components.any?
     iri
--- a/app/serializers/application_serializer.rb
+++ b/app/serializers/application_serializer.rb
@@ -4,7 +4,7 @@ class ApplicationSerializer < ActiveModel::Serializer
   end
 
   def qualified_locid_for(resource, *commands, **query_components)
-    iri = "#{Ontohub::Application.config.fqdn}#{resource.locid}"
+    iri = "#{Hostname.url_authority}#{resource.locid}"
     iri << "///#{commands.join('///')}" if commands.any?
     iri << "?#{query_components.to_query}" if query_components.any?
     iri
--- a/app/serializers/logic_mapping_serializer.rb
+++ b/app/serializers/logic_mapping_serializer.rb
@@ -3,7 +3,9 @@ class LogicMappingSerializer < ApplicationSerializer
     attributes :iri
 
     def iri
-      urls.logic_mapping_url(object, host: Settings.hostname)
+      urls.logic_mapping_url(object,
+                             host: Settings.hostname,
+                             port: Ontohub::Application.config.port)
     end
   end
 
--- a/app/serializers/logic_mapping_serializer.rb
+++ b/app/serializers/logic_mapping_serializer.rb
@@ -4,7 +4,7 @@ class LogicMappingSerializer < ApplicationSerializer
 
     def iri
       urls.logic_mapping_url(object,
-                             host: Settings.hostname,
+                             host: Ontohub::Application.config.fqdn,
                              port: Ontohub::Application.config.port)
     end
   end
--- a/config/initializers/hostname.rb
+++ b/config/initializers/hostname.rb
@@ -6,7 +6,22 @@ module Hostname
     if hostname == 'localhost' || (hostname && hostname.include?('.'))
       hostname
     else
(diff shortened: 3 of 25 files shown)


eugenk:

eugenk:
I rebased this on staging. Tests on travis and jenkins are going to be red because of #1327.

config/initializers/hostname.rb
```diff
@@ -1,16 +1,17 @@
...
 module Hostname
   def self.fqdn
-    hostname = Settings.hostname.split(':').first
-    if hostname == 'localhost' || hostname.include?('.')
+    hostname = Settings.hostname.split(':').first if Settings.hostname
+    if hostname == 'localhost' || (hostname && hostname.include?('.'))
       hostname
     else
-      Addrinfo.tcp(tmp, 0).getnameinfo.first
+      Addrinfo.tcp(Socket.gethostname, 0).getnameinfo.first
```
0robustus1: Shouldn't this fail with a more useful error message than `SocketError: getaddrinfo: nodename nor servname provided, or not known`?

0robustus1:
Ok let's start with the easy thing:
Unless a correct hostname/port-combo is placed inside the settings.yml, this approach won't work for development machines (it will default to localhost without a port, and development-machines don't run ontohub on port 80).

In addition to that, everywhere where you've used the url-rails helpers, and therefore set `host: ...`, you'll also need to set `port: ...`. And for the (qualified) loc/ids i propose another helper method, which actually builds the complete authority (scheme + fqdn + port) in order to use it there, e.g. in lib/hets.rb.

eugenk:
> Unless a correct hostname/port-combo is placed inside the settings.yml, this approach won't work for development machines (it will default to localhost without a port, and development-machines don't run ontohub on port 80).

I don't understand this. The [settings/development.yml](https://github.com/ontohub/ontohub/blob/use_os_hostname_if_hostname_is_undefined_in_settings/config/settings/development.yml#L2) specifies the port 3000.

eugenk:
Maybe the port problem lied in what b941d9b3af35eaf1562109312ce79aad8e2f783b fixed. That commit also should have fixed #1345.

eugenk:
Jenkins, test this please.

eugenk:
Now this should be done.

app/serializers/logic_mapping_serializer.rb
```diff
@@ -3,7 +3,9 @@ class Reference < ApplicationSerializer
     attributes :iri
 
     def iri
-      urls.logic_mapping_url(object, host: Settings.hostname)
+      urls.logic_mapping_url(object,
+                             host: Settings.hostname,
```
0robustus1: Why `Settings.hostname` instead of `Ontohub::Application.config.fqdn`?

app/serializers/application_serializer.rb
```diff
@@ -4,7 +4,7 @@ def urls
   end
 
   def qualified_locid_for(resource, *commands, **query_components)
-    iri = "#{Settings.hostname}#{resource.locid}"
+    iri = "#{Ontohub::Application.config.fqdn}#{resource.locid}"
```
0robustus1: This would be a place for `url_authority`, wouldn't it?

spec/models/ontology_version_spec.rb
```diff
@@ -37,7 +37,7 @@
     context 'in subdirectory' do
       let(:ontology) { create :ontology, basepath: 'subdir/pizza' }
       let(:qualified_locid) do
-        "http://#{Settings.hostname}#{ontology_version.locid}"
+        "http://#{Ontohub::Application.config.fqdn}#{ontology_version.locid}"
```
0robustus1: Shouldn't this be the `url_authority` in order to include the port?

lib/hets.rb
```diff
@@ -17,7 +17,7 @@ def self.minimum_version
 
   def self.qualified_loc_id_for(resource)
     locid = URI.escape(resource.versioned_locid)
-    "http://#{Settings.hostname}#{locid}"
+    "#{Hostname.url_authority}#{locid}"
```
0robustus1: Is `Ontohub::Application.config` not defined here?

config/initializers/hostname.rb
```diff
@@ -1,16 +1,38 @@
...
 
   def self.port
-    Settings.hostname.split(':').last
+    Settings.hostname.split(':').last if Settings.hostname.include?(':')
+  end
+
+  def self.url_authority(scheme: 'http')
+    port = Ontohub::Application.config.port
+    port = ":#{port}" if port
+    "#{scheme}://#{Ontohub::Application.config.fqdn}#{port}/"
```
0robustus1: The trailing slash should break every usage with loc-ids (which have a leading slash themselves).

lib/hets.rb
eugenk: It is defined there, but the `url_authority` is a method and not an attribute of `config` ;)

eugenk:
Good catches. Looks like I forgot how to grep.

0robustus1:
Ran seeds and tests, seems ok now: :+1:

PR #1313, opened 2015-04-10
Fix time taken being negative.

eugenk:
This shall fix the "-1 seconds" at "time taken" bug. The `-1` is a value we get from hets so we have no influence on how it is computed. It seems to only occur on instantaneously returning eprover.

### DIFF
--- a/lib/hets/prove/prove_evaluation_helper.rb
+++ b/lib/hets/prove/prove_evaluation_helper.rb
@@ -22,7 +22,7 @@ module Hets
         proof_attempt.proof_status = find_proof_status_with_hash(proof_info)
         proof_attempt.prover = find_or_create_prover_with_hash(proof_info)
         proof_attempt.prover_output = proof_info[:prover_output]
-        proof_attempt.time_taken = proof_info[:time_taken]
+        proof_attempt.time_taken = time_taken_from_hash(proof_info)
         proof_attempt.tactic_script = tactic_script_from_hash(proof_info)
         used_axioms, used_theorems, generated_axioms =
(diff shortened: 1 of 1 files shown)


lib/hets/prove/prove_evaluation_helper.rb
```diff
@@ -22,7 +22,7 @@ def fill_proof_attempt_instance(proof_attempt, proof_info)
         proof_attempt.proof_status = find_proof_status_with_hash(proof_info)
         proof_attempt.prover = find_or_create_prover_with_hash(proof_info)
         proof_attempt.prover_output = proof_info[:prover_output]
-        proof_attempt.time_taken = proof_info[:time_taken]
+        proof_attempt.time_taken = time_taken_from_hash(proof_info)
         proof_attempt.tactic_script = tactic_script_from_hash(proof_info)
```
0robustus1: Why did you decide to send the hash to the method? I'm just interested in your reasoning.

eugenk: That's just to be uniform with all the other methods extracting information. I think reading the calling method is easier if you always send the same parameter and it doesn't really hurt the readability of the called method.

0robustus1: Alright, but then you should probably call all these methods `*from_hash` and not mix it with `*with_hash`.

eugenk: Done with bdc6dc5

tillmo:
:+1:

PR #1315, opened 2015-04-10
1276 improve selection of provers

### LINKED ISSUE #1276: Improve selection of provers view
Based on #1237: it would be nice to have a "select all" button. Also, if there are many provers, the layout could use several columns (more compact).

eugenk:
This shall fix #1276.


tillmo:
:+1:

PR #1316, opened 2015-04-11
Fix logicgraph generation

eugenk:
The previous code worked flawlessly on a mac, but my ubuntu system could not execute `system` properly. `system` returned `nil` with `$?` being `nil` as well. With the backtick notation it works.


eugenk:
As suspected by @0robustus1, the shell spawning syntax of `system` works just fine.

0robustus1:
:+1:

PR #1304, opened 2015-04-04
1156 add combinations plus api integration

0robustus1:
Shall close #1156.

This adds the basic Creation of combined ontologies as a feature. Currently only nodes (ontologies) can be combined. But this can be later improved upon.

PR #1317, opened 2015-04-11
1305 add actions support

### LINKED ISSUE #1305: Add /actions for API
In order to support asynchronous operation in a way that is actually compatible with REST,
we will introduce the '/actions' resource. The cycle for a long-running operation would be like this:

- POST for long-running operation
  - respond with 202 Accepted and a `Location:` header to a `/actions/:id`
- GET to `/actions/:id`
  - respond with 200 OK and a JSON-body with a `status`, `eta` (seconds), and if status is terminal a reference to the resulting resource.

0robustus1:
Seems i forgot to issue a pull-request for this. It shall close #1305. It is based on the #1156-er branch,
so #1304 should be reviewed first.


0robustus1:
#1304 is merged, so i rebased staging.

spec/models/action_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+      expect(action.eta(target + 5.minutes)).to eq(0)
+    end
+  end
+
+  context '#status' do
+    let(:action) { create :action, resource: resource }
+    let(:status) { 'some status' }
+
+    context 'on resource with state' do
+      let(:resource) { mock_model("Resource", state: status) }
```
eugenk: Please use single quotes (in the lines below, too).

app/models/action.rb
```diff
@@ -0,0 +1,38 @@
...
+  attr_accessible :initial_eta, :resource
+
+  def self.enclose!(initial_eta, klass, method, *args)
+    action = create!(initial_eta: initial_eta)
+    ActionWorker.perform_async(action.id, klass.to_s, method, *args)
+    action
+  end
+
+  def eta(time = Time.now)
+    diff = (created_at + initial_eta) - time
```
eugenk: I'm just beginning to read the changes, so I don't know about the mechanics, but here's what popped into my mind:

If sidekiq is involved and the actiom model is created synchronously, the job could not start until after the eta is reached. Maybe we should use `updated_at` and `touch` the action as soon as we start processing the job.

eugenk:
Looks good. What do you think of `created_at` vs `updated_at` in the `eta`?

0robustus1:
@eugenk I've partially adopted your `updated_at` solution. The latest commits add a `action_started_at` field, which is touched when the `ActionWorker` does its thing.

eugenk:
Instead of `action_started_at` I slightly prefer `started_at`, however, this is no deal breaker. If you don't agree, you can leave the name as is. Having said that I give you :+1:

PR #1320, opened 2015-04-12
Fix run seeds when offline

eugenk:
Seeds depended on an internet connection. With this change they don't:
* Proof statuses and categories have fixtures. They are only used if the download from ontohub.org failed.
* Oops is skipped if either a SocketError (no connection at all) or a RestClient error (e.g. 404 or similar) occurs.

In both cases a message with the error is printed to stdout.

### DIFF
--- a/db/seeds/070-oops.rb
+++ b/db/seeds/070-oops.rb
@@ -1,9 +1,9 @@
 def handle_error(e, reason)
-  puts <<MSG
-  Could not seed "oops" because #{reason}.
-  #{e.class}: #{e.message}
-  Continuing with other data.
-MSG
+  puts <<-MSG.squish
+    Could not seed "oops" because #{reason}.
+    #{e.class}: #{e.message}
+    Continuing with other data.
+  MSG
 end
 
 # Add OOPS! requests and responses to pizza ontology.
--- a/db/seeds/070-oops.rb
+++ b/db/seeds/070-oops.rb
@@ -1,18 +1,32 @@
+def handle_error(e, reason)
+  puts <<MSG
+  Could not seed "oops" because #{reason}.
+  #{e.class}: #{e.message}
+  Continuing with other data.
+MSG
+end
+
 # Add OOPS! requests and responses to pizza ontology.
 ontology = Ontology.where(name: "Pizza").first!
 version  = ontology.versions.first
-if version
-  request  = version.build_request({state: 'done'}, without_protection: true)
-  request.save!
+begin
+  if version
+    request  = version.build_request({state: 'done'}, without_protection: true)
+    request.save!
 
-  responses = %w(Pitfall Warning Warning Suggestion).map do |type|
-    request.responses.create! \
-    name:         Faker::Name.name,
-    code:         0,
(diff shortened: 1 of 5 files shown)


eugenk:
As requested in the meeting yesterday, I changed the default behaviour to use the local fixtures only. By setting an environment variable (`DOWNLOAD_FIXTURES`), we can choose to use online files from ontohub.org instead. The rebuild script sets this environment variable if the `-d` option is supplied.

Also, `-i` makes the rebuild script only restart invoker and not clean the state.

0robustus1:
I'd prefer `-r` (for `restart`) instead of `-i` (for `invoker`, i assume?).

eugenk:
Renamed with 51bed36

db/seeds/070-oops.rb
```diff
@@ -1,18 +1,32 @@
+def handle_error(e, reason)
+  puts <<MSG
```
0robustus1: Please use the `<<-MSG` way and indent the closing `MSG` correctly.

db/seeds/070-oops.rb
```diff
@@ -1,18 +1,32 @@
...
+  Could not seed "oops" because #{reason}.
```
0robustus1: Indent to the far left corner, or use `squish`.

lib/tasks/generate.rake
```diff
@@ -16,13 +16,22 @@ end
...
   basename = File.basename(filepath)
+  if ENV['DOWNLOAD_FIXTURES']
+    source_url =
+      "https://ontohub.org/repositories/meta/master/download/#{source_file}"
+    temp_file = Tempfile.new([target_file, File.extname(target_file)]).path
+    if system("wget -O #{temp_file} #{source_url}")
+      filepath = temp_file
+    else
+      puts
+        'No connection to ontohub.org. Using local file for the current task.'
```
0robustus1: That won't put the string, but actually prints an empty line (`puts` can take zero arguments).
Either put the string in the same line as `puts` or use Docstrings.

script/rebuild-ontohub
```diff
@@ -34,7 +34,26 @@ execute_or_die() {
 abort_unless_invoker_installed
 initialize_new_invoker_instance
 
-execute_or_die "bundle exec rake elasticsearch:wipe"
-execute_or_die "bundle exec rake db:migrate:clean"
-execute_or_die "redis-cli flushdb"
-execute_or_die "bundle exec rake db:seed"
+for i in "$@"
```
0robustus1: please use `; do` ;)

script/rebuild-ontohub
```diff
@@ -34,7 +34,26 @@ execute_or_die() {
...
+    RESTART_INVOKER_ONLY=true
+    ;;
+    *)
+    # unknown option
+    ;;
+  esac
+  shift
+done
+
+if [[ !($RESTART_INVOKER_ONLY) ]]
```
0robustus1: please use `; then`

0robustus1:
:+1:

PR #1322, opened 2015-04-15
Report coverage only on default task

eugenk:
A simple run of `rspec spec/something_spec.rb:123` spends a lot of time on
generating the coverage report.

This disables the coverage report, except for the default rake task.
If a coverage report is desired even for other tasks, just set the environment variable `COVERAGE`, e.g.:

```
COVERAGE=true rspec
```

### DIFF
--- a/spec/shared_helper.rb
+++ b/spec/shared_helper.rb
@@ -34,7 +34,6 @@ module SharedHelper
   end
 
   def use_simplecov
-    return unless ENV['COVERAGE']
     SimpleCov.start do
       add_group "Models",      "app/models"
       add_group "Controllers", "app/controllers"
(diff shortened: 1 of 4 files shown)


spec/shared_helper.rb
```diff
@@ -34,6 +34,7 @@ def gemsets
   end
 
   def use_simplecov
+    return unless ENV['COVERAGE']
```
0robustus1: Shouldn't the caller of `use_simplecov` decide whether simplecov is supposed to be used? If not, we should rename the method.

0robustus1:
Please rebase staging.

spec/shared_helper.rb
eugenk: Yes, you're right.

eugenk:
Done.

0robustus1:
:+1:

PR #1323, opened 2015-04-16
Remove obsolete database cleaner config

eugenk:
After moving all MiniTests to RSpec, we don't need the config for MiniTest anymore.


0robustus1:
:+1:

PR #1327, opened 2015-04-23
1325 parse hets json snake case keys 

eugenk:
This shall fix #1325 and also use the new `display_name` of Hets' provers list.

Hets now outputs JSON with snake case (instead of camel case) object keys. Our parsers need to be adjusted to this.

--
*Because our CI services use the bleeding-edge version of Hets, they will always fail without the changes of this branch.*

PR #1330, opened 2015-04-23
Fix ssh key fingerprint generation

### LINKED ISSUE #1329: NotNullViolation when adding ssh key
When I try to add a ssh key I get 
```

ActiveRecord::StatementInvalid at /keys

PG::NotNullViolation: ERROR:  null value in column "fingerprint" violates not-null constraint
DETAIL:  Failing row contains (1, 1, ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAAB/Rferei3OhJ/JhKOm3thTdMkFpTP..., test, null, 2015-04-23 09:56:34.929769, 2015-04-23 09:56:34.929769).
: INSERT INTO "keys" ("created_at", "fingerprint", "key", "name", "updated_at", "user_id") VALUES ($1, $2, $3, $4, $5, $6) RETURNING "id"
```
The error appears in line 285 in `lib/active_record/connection_adapters/abstract_adapter.rb`

eugenk:
This shall fix #1329.

Openssh changed the fingerprint output format in version 6.8. Now it
defaults to SHA256 (previously MD5) and it prepends `MD5:` or `SHA256:`
to the fingerprint.

With `-E md5` we tell it to use the old fingerprint format. This option
is not recognized by versions before 6.8 and calling `ssh-keygen` with
it lets the process fail. To support the older versions as well
(they are running on our productive deployments), we call it a second
time with correct options if we find ourselves in in an environment with
the newer version.

### DIFF
--- a/app/models/key/fingerprint.rb
+++ b/app/models/key/fingerprint.rb
@@ -3,7 +3,7 @@ module Key::Fingerprint
 
   included do
     before_validation :generate_fingerpint
-    validates :fingerprint, uniqueness: true
+    validates :fingerprint, uniqueness: true, presence: true
   end
 
 
--- a/app/models/key/fingerprint.rb
+++ b/app/models/key/fingerprint.rb
@@ -16,6 +16,15 @@ module Key::Fingerprint
       file.puts key
       file.rewind
       output = Subprocess.run 'ssh-keygen', '-lf', file.path
+      # Openssh 6.8 changed the format and hash algorithm of the fingerprint.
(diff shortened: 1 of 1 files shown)


eugenk:
One reason for our test suites to fail is that #1327 is not yet merged.

eugenk:
I suggest that you run the tests locally and ignore Travis and Jenkins. As long as you don't have the latest Hets version this is the way to go.

ebolloff:
:+1:

PR #1331, opened 2015-04-24
Refactor history spec

eugenk:
The history spec had poorly chosen `describe` and `context` descriptions, which are changed here.

Also the nesting of the contexts was bad because there were two `describe` blocks. I changed them into two `context`s inside of one `describe`.

The diff is only this huge because the indentation changed.


0robustus1:
:+1:

PR #1332, opened 2015-04-24
Refactoring get rid of rspec deprecations

eugenk:
This removes deprecated method calls in our specs. It only changes what is not already done in #1317. When #1317 and this PR are merged, rspec should no longer print deprecation warnings.


0robustus1:
:+1:

PR #1333, opened 2015-04-25
Use less verbose test output formatter

eugenk:
The quite verbose formatter annoyed me a little because it printed way more information than we needed. This pull request adds formatters that combine the strengths of the very verbose formatters `documentation` (for rspec) and `pretty` (for cucumber) and of the too little printing formatter `progress`.

Those are only active by default on development machines. I made sure that the verbose formatters are used on travis and jenkins.

#### rspec
screen shot 2015-04-25 at 11 12 33

#### cucumber
screen shot 2015-04-25 at 11 13 34

### DIFF
--- a/.rspec
+++ b/.rspec
@@ -1,2 +1,3 @@
 --color
---format documentation
+--require ./spec/support/documentation_progress_formatter.rb
+--format DocumentationProgressFormatter
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ before_install:
   - 'gem install bundler'
 
 script:
-  - SPEC_OPTS="--format documentation" ELASTIC_TEST_PORT=9200 bundle exec rake
+  - SPEC_OPTS='--format documentation' CUCUMBER_OPTS='--format pretty' ELASTIC_TEST_PORT=9200 bundle exec rake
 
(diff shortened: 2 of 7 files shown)


0robustus1:
Mhm, per default i would only activate the middleground for rspec (where it is really necessary). For cucumber i think that the verbose output is actually necessary/extremely useful.

eugenk:
I think verbose output in cucumber is only useful when something fails. Otherwise it's just some green text that you don't read anyway. At least I don't.

In my opinion it clutters the console. We don't have that many cucumber features at this point, but when it gets more, I think we really want to shorten the output.

0robustus1:
Yeah, but you do not get the full output for a failing test, and i think for cucumber (with their step-system) it is important to see them all.

0robustus1:
@ebolloff want to be the tie-breaker?

eugenk:
Yes, then I think I should change the formatter to print the full output of a failing test.

0robustus1:
If that would be possible, it would probably be the optimal case.

eugenk:
How about now? "a44c794 Print failure details after the features." should add the details.

0robustus1:
:+1:

PR #1334, opened 2015-04-27
quote result of subprocess

### LINKED ISSUE #1328: Broken pipe in rebuild-ontohub script
There's a non-deterministic error `Errno::EPIPE: Broken pipe` when I do `script/rebuildontohub` on branch `1170-git_handling`. The error appears at the seeds `070-oops.rb`.

0robustus1:
Shall close #1328.

Otherwise the result is executed inside of
the bash script, which is problematic if it
contains semicoli for example.

### DIFF
--- a/script/rebuild-ontohub
+++ b/script/rebuild-ontohub
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 abort_unless_invoker_installed() {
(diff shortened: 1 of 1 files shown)


eugenk:
:+1:

PR #1335, opened 2015-04-29
Fix data migrations

eugenk:
This moves some data-changing migrations from `db/data/` to `db/migrate/` because
* executing them shouldn't take too long, so they can be executed synchronously
* they don't need complex methods
* they are needed for ontohub to work properly

The migrations have been adjusted to directly work on the database and to not depend on model methods. This adds some database queries, but they are only called once so the extra execution time shouldn't hurt.


eugenk:
I don't know who to assign to this pull request. This is connected to both, ontohub in general and the reasoning milestone of my master's thesis.

eugenk:
To test the migrations, I recommend to
1. Check out 7e216f85f8d860cc3948b656c0d0066ccf0dafbb - This has proving functionality, but is before all the moved migrations.
2. `rake db:migrate:clean` and `rake db:seed`
3. **Prove something** - Proofs are not in the seeds in this commit.
4. Check out this branch
5. Migrate

tillmo:
Sometimes code is duplicated from models (or lib) into the migrations in order to ensure that they work independently of whether this code is present. Looks a bit odd, but seems to be necessary. :+1:

PR #1339, opened 2015-05-05
Use correct table for JOIN

### LINKED ISSUE #1338: column oops_responses_symbols.id does not exist
When opening the symbols view of Pizza (http://ontohub.dev/default/pizza///symbols?kind=Class), we get:

```
!! #<ActiveRecord::StatementInvalid: PG::UndefinedColumn: ERROR:  column oops_responses_symbols.id does not exist
LINE 1: ...ps_responses" LEFT JOIN oops_responses_symbols ON oops_respo...
                                                             ^
: SELECT "oops_responses".* FROM "oops_responses" LEFT JOIN oops_responses_symbols ON oops_responses_symbols.id = oops_responses_symbols.oops_response_id WHERE "oops_responses"."oops_request_id" = 1 AND "oops_responses_symbols"."symbol_id" IS NULL>
```

eugenk:
Shall fix #1338.

When introducing `table`, we accidentally replaced the wrong table name
with the `table` variable.

### DIFF
--- a/app/models/oops_response.rb
+++ b/app/models/oops_response.rb
@@ -4,7 +4,7 @@ class OopsResponse < ActiveRecord::Base
 
   scope :global, ->() do
     table = 'oops_responses_symbols'
-    joins("LEFT JOIN #{table} ON #{table}.id = #{table}.oops_response_id").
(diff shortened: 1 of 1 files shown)


0robustus1:
:+1:

PR #1357, opened 2015-05-27
necessary changes to make ubuntu developer installation instructions work

tillmo:
@CarloDib has tried out the ubuntu developer installation and failed. This corrects some problems he encountered.

### DIFF
--- a/doc/developer_installation_on_ubuntu.md
+++ b/doc/developer_installation_on_ubuntu.md
@@ -71,17 +71,11 @@ After this we should create the config directory for postgres:
 
 - `initdb /usr/local/var/postgres -E utf8`
 
-Now when you use `psql` you should be connected to the
-database (`\q` lets you quit).
+Now when you use
+- `sudo su -l postgres`
+- `psql` 
+you should be connected to the database (`\q` lets you quit).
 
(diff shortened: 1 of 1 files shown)


doc/developer_installation_on_ubuntu.md
```diff
@@ -92,8 +86,16 @@ on the system, so we will have to install one:
 
 - `sudo apt-get install libqtwebkit-dev`
 
+Clone the ontohub sources:
+- `git clone git@github.com:ontohub/ontohub.git`
+Note that you need to get access rights from the Ontohub team first.
```
0robustus1: Actually a clone via ssh/git still works if you do not have write-permissions on a repository. Only the push will later fail (as it should).

tillmo: good point, the I will change this to ssh/git only.

tillmo:
btw: do you know why postgres in installed from a ppa, and not from the regular Ubuntu sources?

0robustus1:
At the time of writing the version in the ubuntu sources did not contain features we were using. I'm not sure if that has changed.

Wrt. to the Pull Request: :+1:

PR #1359, opened 2015-05-31
Fix and allow exceptions for rebuild ontohub

0robustus1:
Shall improve the rebuild-ontohub script.

- Execution does not cease if the elasticsearch-wiper gets a 404 response (which basically means that the cache was already wiped)
- Adds the new command to allow such special cases
- Adds the `--no-invoker` switch, which will result in the script doing the following
  - no check if invoker is installed
  - no start of invoker
  - no restart of invoker

### DIFF
--- a/script/rebuild-ontohub
+++ b/script/rebuild-ontohub
@@ -58,8 +58,7 @@ abort_cmd() {
   exit 1
 }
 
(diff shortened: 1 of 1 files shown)


ebolloff:
:+1:

PR #1367, opened 2015-06-04
Fix: Bundle is not complete for deployment

### LINKED ISSUE #1366: Bundle is not complete for deployment
VCR is used in the rake tasks, but it is only in the `test` group of the Gemfile. As soon as the rake tasks are loaded during deployment we get an error.

eugenk:
Shall fix #1366.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -118,6 +118,10 @@ gem 'dagnabit', '~> 3.0.1'
 # Migrate data in separate tasks
 gem 'data_migrate', '~> 1.2.0'
 
+# Recording of HTTP Requests
+gem "vcr", '~> 2.9.2'
+gem "webmock", '~> 1.9.0'
+
 group :test do
   gem 'mocha', '~> 1.1.0', require: false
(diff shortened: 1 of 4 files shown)


eugenk:
I cherry-picked this commit to develop.ontohub.org (with conflict solving) and it works there.

0robustus1:
Could we document this somewhere? (What files/fixtures are used and why)

eugenk:
You mean to extend [doc/most_important_gems.md](https://github.com/ontohub/ontohub/blob/20c5bf7fe7a97c1f44fef88424cfd8619735feb7/doc/most_important_gems.md)?

0robustus1:
That would be the place to add vcr and webmock, but i also want to document (maybe in productive deployment documentation, why we need the files and what they are used for, and when they are subject to change).

eugenk:
Done with cb99870

0robustus1:
After green tests: :+1:

PR #1375, opened 2015-06-15
1374 Improve migration performance

### LINKED ISSUE #1374: Improve migration performance
Working on #1361, we found that some migrations consume way too much memory, especially `db/migrate/20150311125212_make_sentences_to_axioms.rb`.

Use, for example, `MyModel.select(:id).find_each` or something like
```ruby
def find_each_with_garbage_collection(options = {}, &block)
  find_in_batches(options) do |group|
    group.each { |record| block.call(record) }
    GC.start
  end
end
```
to improve the performance.

eugenk:
This shall fix #1374. I tried to use the garbage collection invoking method, but that's slower than simply restricting the select statement to the one needed column.

The other migrations don't seem to be problematic on the ontohub.org data.


0robustus1:
:+1:

Issue #1351, opened 2015-05-24
Many quick edits lead to "This iri <...> is already being processed"


tillmo:
When I quickly edit an ontology file several times through the web interface, this leads to "This iri <...> is already being processed". Actually, I would expect that no error occurs and (at least) the latest version is being processed. This means that instead of throwing the error "This iri <...> is already being processed", the job should wait until the other job has been finished.

tillmo:
I get the error even when uploading a new version *after* the first version has failed, see http://staging.ontohub.org/test-um/Chapter8///ontology_versions

tillmo:
I even get the error for the first version of a file: http://staging.ontohub.org/test-um/Chapter8V2///ontology_versions

tillmo:
possibly this is related to the fact that Hets-lib libraries are imported, which are treated specially. (Then, this special treatment would need to be revised)

0robustus1:
Ok, well the issue doesn't seem to be on the ontohub side at all. It seems to be entirely a hets-issue:

```
<DGNode name="" reference="false" logic="CASL" refname="?" relxpath="/">
    <Symbols>
      <Symbol kind="sort" name="Nat" iri="Nat" range="/usr/local/opt/hets-lib/UserManual/Chapter3.casl:167.15-167.17">sort Nat</Symbol>
      <Symbol kind="op" name="0" iri="0" range="/usr/local/opt/hets-lib/UserManual/Chapter3.casl:167.23">op 0 : Nat</Symbol>
      <Symbol kind="pred" name="__&lt;__" iri="__&lt;__" range="/usr/local/opt/hets-lib/UserManual/Chapter3.casl:185.12-185.16">pred __&lt;__ : Nat * Nat</Symbol>
      <Symbol kind="op" name="suc" iri="suc" range="/usr/local/opt/hets-lib/UserManual/Chapter3.casl:167.27-167.29">op suc : Nat -&gt; Nat</Symbol>
    </Symbols>
  </DGNode>
  <DGNode name="" reference="false" logic="CASL" refname="?" relxpath="/">
    <Symbols>
      <Symbol kind="sort" name="C
```

As we determine IRIs (and loc/ids) by the DGNode name attribute we expect the `name` to be unique inside of one hets-xml DGraph output.

0robustus1:
I've created spechub/hets#1503 to tackle this problem on the hets-side.

PR #1378, opened 2015-06-19
1351 remove iri field

0robustus1:
During the course of #1351 i've removed the iri-field from ontologies in ontohub. They are superseded by the locid-field. The iri is now only a method to generate the fully-qualified locid.


eugenk:
Looks good to me. :+1:

PR #1385, opened 2015-06-26
Fix elasticsearch docs: Mention path to config file.

eugenk:
The docs were lacking the path to the config file.


0robustus1:
:+1:

PR #1387, opened 2015-06-26
Fix axiom selection

eugenk:
This fixes a bug that made the axiom selection a no-op.

### DIFF
--- a/spec/lib/proof_spec.rb
+++ b/spec/lib/proof_spec.rb
@@ -88,7 +88,7 @@ describe Proof do
       end
     end
 
-    context 'axiom selection is correct' do
+    context 'axiom selection' do
       let(:params_modified) do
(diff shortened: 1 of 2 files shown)


spec/lib/proof_spec.rb
```diff
@@ -87,6 +88,28 @@
...
+        end
+      end
+
+      it 'right inclusion' do
+        proof_modified.instance_variable_get(:@axioms).each do |selection|
+          expect(axioms.map(&:name)).to include(selection)
+        end
+      end
+    end
+
```
tillmo: what about using match_array here?

eugenk: Thanks! I didn't know that matcher. It's fixed in e688e7b

tillmo:
:+1:

PR #1389, opened 2015-06-30
Explicitly set IDs on form inputs

### LINKED ISSUE #1388: Invalid HTML
There are two forms on the startpage, one for Sign In and one for Sign Up. The input fields e-mail and password have the same id `user_email` and `user_password` in both forms.

eugenk:
Shall fix #1388.


ebolloff:
:+1:

PR #1390, opened 2015-06-30
Suppress Rspec verbose output

eugenk:
At the beginning of each run of `rake`, the long `db:migrate:clean` output as well as the list of the spec files annoyed me a little. This change hides that output. The important things, such as the spec progression and the spec report are still shown.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -7,6 +7,13 @@ require File.expand_path('../config/application', __FILE__)
 
 Ontohub::Application.load_tasks
 
+# Prevent Rspec to print file list unless enforced by the environment.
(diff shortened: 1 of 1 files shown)


eugenk:
I just rebased on staging (after merging #1392). Hopefully Travis is green now.

ebolloff:
One time the output starts with `Repository destroying ..` and the other time with `TeamsController`. Why is that?

eugenk:
The specs are executed in a random order. This increases the probability to find erroneous specs (dependent on other specs to run first).

ebolloff:
Learning never stops. :+1:

PR #1392, opened 2015-07-01
Temporarily set libv8 to a previous version

eugenk:
The current version is not installable on several Linux clients, including travis.


ebolloff:
Works on my machine, hopefully on others, too.  :+1:

PR #1403, opened 2015-07-03
1398 Add secret_token yml-key to settings

### LINKED ISSUE #1398: Add secret_token yml-key to settings
@jelmd wants the `secret_token` to be in the yml files so that he needs to touch fewer files for the configuration. I don't see a problem in putting the `secret_token` into the `config/settings.local.yml` as long as this file is not publicly available, but maybe I didn't think of everything.

I suggest to put something like this into the `settings.yml`:
```yml
# Your secret key for verifying the integrity of signed cookies.
# If you change this key, all old signed cookies will become invalid!
# Make sure the secret is at least 64 characters and all random,
# no regular words or you'll be exposed to dictionary attacks.
secret_token: # nil - specify this key in the settings.local.yml and keep it secret
```
And this in the `config/initializers/secret_token.rb`:
```ruby
Ontohub::Application.config.secret_token = Settings.secret_token
```

Next, I'm not sure, if maybe the `config/secrets.yml` is the right place for it, but that would not reduce the number of config files that need to be touched...

@SGraef, @ebolloff, @tillmo (and maybe @0robustus1 if you still have time for a comment), what are your thoughts on this?

eugenk:
This shall fix #1398

### DIFF
--- a/config/initializers/secret_token.rb
+++ b/config/initializers/secret_token.rb
@@ -4,4 +4,4 @@
 # If you change this key, all old signed cookies will become invalid!
 # Make sure the secret is at least 30 characters and all random,
 # no regular words or you'll be exposed to dictionary attacks.
-Ontohub::Application.config.secret_token = '6b198bfe3759ee41524d3a0d7547890a2d277f44f2ce472921ced9bac1833830028bbcacf58fdac0b482265042baa8df4de86a86ad63711ef5b50f70e57d5a07'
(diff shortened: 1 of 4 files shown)


SGraef:
The rake task secret:replace must be changed too

eugenk:
done with  1672913

SGraef:
:+1:

PR #1404, opened 2015-07-05
Fix settings validations for the tests

eugenk:
Whenever the settings validations failed, travis and jenkins were green.
However, the settings validations fail because of missing directories. Those are now created in both, travis and jenkins (see also https://github.com/ontohub/ontohub-ci/commit/b7e0d081c62210c8b2712626ee01cc8634650eca).


SGraef:
:+1:

PR #1405, opened 2015-07-06
Update gems and add puma

eugenk:
This shall
* resolve the redcarpet security issue (CVE-2015-5147)
* update some other gems - including rails-erd which has been outdated for a while and finally can be updated because ruby-graphviz got an update as well
* fix https://github.com/ontohub/ontohub/issues/1170#issuecomment-118408150 by adding puma to the production gems


SGraef:
:+1:

PR #1408, opened 2015-07-07
1407 Allow to set an email without the host-part

eugenk:
Now we can either set `noreply` or `noreply@example.com` as the email
address. If the host-part is omitted, it will be appended automatically.
If the host-part is specified, it must be the same as the FQDN.


SGraef:
:+1:

Issue #1395, opened 2015-07-01
problem with e-mail config


jelmd:
sending email does not work:
```
ActionView::Template::Error (undefined method `merge' for #<RailsConfig::Options host="ta.iws.cs.ovgu.de", port=nil>):
    1: %p
    2:   Hello #{@resource.email}!
    3: %p Someone has requested a link to change your password, and you can do this through the link below.
    4: %p= link_to 'Change my password', edit_password_url(@resource, reset_password_token: @token)
    5: %p If you didn't request this, please ignore this email.
    6: %p Your password won't change until you access the link above and create a new one.
  app/views/devise/mailer/reset_password_instructions.html.haml:4:in `_app_views_devise_mailer_reset_password_instructions_html_haml___1152300800981300088_70053456957420'
```

see settings on ta

eugenk:
It seems like a problem with some gems. 

`ruby/2.1.0/gems/actionpack-3.2.22/lib/action_dispatch/routing/routes_proxy.rb:15` returns a `RailsConfig` object on ta, while it returns a simple `Hash` on develop.ontohub.

This might be a result of having a different bundle on ta than on the other machines.

I'd like to try to use the exact same bundle on ta (i.e. the unpatched Gemfile and Gemfile.lock), but there is a problem: The commit we have in the Gemfile.lock for rails_config doesn't exist any more (the referenced branch was rebased). Fortunately, I still have that commit in my own fork. So if we change
```
remote: git://github.com/hnakamur/rails_config.git
``` 
to
```
remote: git://github.com/eugenk/rails_config.git
```
in the Gemfile.lock, and
```
gem 'rails_config', github: 'hnakamur/rails_config', ref: 'overwrite_arrays'
```
to
```
gem 'rails_config', github: 'eugenk/rails_config', ref: 'overwrite_arrays'
```
in the Gemfile, we can get the same bundle.

I don't know how you installed the bundle, @jelmd, so could you please take care of that and in the same step not add puma to the Gemfile (at least for this investigation).

I suspect (but can't say for sure) that the latest changes in rails_config introduced this problem. They have been added to the referenced branch by rebase.

jelmd:
To see, what the `~/etc/oh-update.sh` does, use `~/etc/oh-update.sh -h` . To enable the debug output for all functions use e.g. `-T ALL`.

So just calling `~/etc/oh-update.sh` is basically a 
* `bundler install --path=/local/home/ontohub --without development test --deployment --jobs=32`
[...]

PR #1416, opened 2015-07-09
Lots of deployment fixes

### LINKED ISSUE #1414: Put cp_keys to data/.ssh/cp_keys
Apparently the `cp_keys` executable is not accessible at `${paths.git_home}/.ssh/cp_keys`, so put it to `${paths.data}/.ssh/cp_keys`

### LINKED ISSUE #1415: Implicit path settings fail
When the paths settings are nil/set implicitly *and* the directories actually exist, the validations fail.

eugenk:
This shall
* fix #1414
* fix #1415
* print a more useful error message if the cp_keys executable is missing. Previously, this printed a bad key and a bad value (only a good error message itself)

### DIFF
--- a/config/initializers/paths.rb
+++ b/config/initializers/paths.rb
@@ -1,8 +1,7 @@
 module PathsInitializer
   DEFAULT_PATHS = {git_repositories: 'repositories',
                    symlinks: 'git_daemon',
-                   commits: 'commits',
-                   git_home: 'git'}
+                   commits: 'commits'}
   class << self
     def expand(path)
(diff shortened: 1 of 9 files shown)


SGraef:
:+1: looks fine to me

eugenk:
Wow, you're fast :o

tillmo:
does this mean that @jelmd can now have another try?

eugenk:
Yes, that would be great. As soon as the next deployment attempt is made, I will investigate #1395 further.

PR #1417, opened 2015-07-09
1395 fix problem with e-mail config

eugenk:
The object in the settings is a RailsConfig. Devise needs it to be a Hash, though.


SGraef:
:+1:

PR #1418, opened 2015-07-09
Set default log level to warn in production mode.

eugenk:
Let's avoid patches and superfluous config files by using better defaults :)


SGraef:
:+1:

PR #1424, opened 2015-07-14
1422 hide prove this theorem button if not authorised

### LINKED ISSUE #1422: Hide "prove this theorem" button if not authorised
On the proof attempts page, if you are not logged in, you still see the "prove this theorem" button. However, if you press it, you get a message that you are not authorised. So the button should not be shown, or perhaps better, greyed out and disabled.

eugenk:
This shall fix #1422.


SGraef:
:+1:

PR #1425, opened 2015-07-14
1024 sentence display broken

### LINKED ISSUE #1024: sentence display broken
In https://github.com/ontohub/ontohub/blob/staging/app/helpers/sentence_helper.rb#L5 we make text strong between `<` and `>`. However, this can break casl sentences like ` x < y /\ y < z => x <z` in `partial_order.casl?strict_partial_order` because the brackets are removed.

eugenk:
This shall fix #1024. It now only displays bold text (removing arrow brackets) for OWL. CASL sentences, for example, still keep their "less than" and "greater than" symbols.

Without this variable being set, the rake task `git:compile_cp_keys` (called by the rebuild script) fails.

### DIFF
--- a/app/helpers/sentence_helper.rb
+++ b/app/helpers/sentence_helper.rb
@@ -1,7 +1,7 @@
 module SentenceHelper
 
   def format_for_view(sentence)
-    if sentence.display_text? then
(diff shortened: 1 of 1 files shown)


SGraef:
:+1:

PR #1426, opened 2015-07-16
Specify GIT_HOME environment variable for rebuild script

eugenk:
The GIT_HOME environment variable was not set in the rebuild script.
This adds a default value.

### DIFF
--- a/doc/developer_installation_on_mac_os_x_lion_and_up.md
+++ b/doc/developer_installation_on_mac_os_x_lion_and_up.md
@@ -154,6 +154,17 @@ brew install hets --with-wrapper
 
 If there is a new version released you can update hets through the homebrew process (`brew upgrade`).
 
+## Extra executables
+
+The SSH-Key handling requires an additional executable `data/.ssh/cp_keys` to exist.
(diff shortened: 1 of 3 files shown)


ebolloff:
There are some things which are broken on my machine right now, thats why I got a red test (which is also red on other branches). Maybe @SGraef wants to check it, too? On my side :+1:

eugenk:
I added another fix because the first commit didn't really use the absolute file path.

0robustus1:
Just a question from my side: Why do we need a specific Environment Variable to be set (mandatory)? Usually those are only used to provide options and it should work with reasonable defaults without specifiying an Env-Var.

eugenk:
There's no place in the ontohub code/settings files any more where we have this path. I could've hard coded the default default directly into the rake task. I didn't do so because many rake tasks have mandatory environment variable settings and I didn't want production systems to have a default.

0robustus1:
Alright. Even though you have already documented the variable in the productive-deployment documentation i would suggest to also document it in development related documentation, just so we can easily know about it in a few months/years time.

eugenk:
Right. Done with d492d49

SGraef:
:+1:

PR #1431, opened 2015-07-20
Overwrite arrays in config

eugenk:
The previous rails_config version didn't properly overwrite arrays.
This version does overwrite them.


SGraef:
:+1:

PR #1432, opened 2015-07-27
Deletes whitespace to fix test.

ebolloff:
This should fix

```
Failure/Error: it{ subject.errors[:key].should == ["is not a public key file."] }
       expected: ["is not a public key file."]
            got: ["is not a public key file.\r"] (using ==)
     # ./spec/models/key_spec.rb:34:in `block (4 levels) in <top (required)>'
```


eugenk:
:+1:

PR #1436, opened 2015-07-28
mapping_spec quick fix

eugenk:
The `mapping_spec` depends on ontohub being available. However, with the current server/TLS problems, we need to remove this dependency. Unfortunately, we need an external source (i.e. not on the same ontohub instance that is tested) for the imported fixture file in order for it to be placed in the `ExternalRepository`.

I couldn't find a clean way to do this without depending on a network connection. So instead of depending on Ontohub, we now depend on Github.


SGraef:
:+1:

PR #1437, opened 2015-08-03
Always create AxiomSelections in migration

eugenk:
If a ProofAttemptConfiguration had no selected axioms, no AxiomSelection
was created. This violated a not null constraint on axiom_selection_id.


SGraef:
:+1:

PR #1442, opened 2015-08-06
Remove unused BasicProof model.

### LINKED ISSUE #230: Fix Errors in BasicProofs Definition
The model `BasicProof` currently has a `belongs_to` relation to `:translation`.

This needs to be changed to `:logic_mapping`.

Furthermore a `BasicProof` `belongs_to` a `:sentence`.

Migrations must also be adjusted.

It is possible that Parsing is also not performed correctly (because
we should experience errors otherwise).

eugenk:
This should fix #230 by removing the model `BasicProof`. The model was never used and we have other models for proofs.


tillmo:
:+1:

PR #1446, opened 2015-08-09
1445 run multiple hets instances

### LINKED ISSUE #1445: Run multiple Hets instances
Currently, we have many Sidekiq workers that listen on the `hets` queue, but only one actual instance of Hets.

* Create as many instances of Hets as we have Sidekiq workers for the `hets` queue.
* Register those instances in the database on creation. Maybe clear the `HetsInstance`s prior to registering.
* Allow to configure the URLs of the Hets instances. This replaces the `Settings.hets.instances_count` by an array `Settings.hets.instance_urls` (such that the count is implicit).
* Create a *simple* load balancing mechanism to choose the "best" instance of Hets. It doesn't need to fetch information from Hets whether it is busy or free. This can be made more sophisticated in another issue.
  * Each HetsInstance gets three new attributes: `state`, `state_updated_at` and `queue_size`
  * As soon as a job on the `hets` queue is started, mark the chosen `HetsInstance` as `busy` and decrease the `queue_size` if it was greater than 0.
  * As soon as a job on the `hets` queue is finished, mark the chosen `HetsInstance` as `free`.
  * When a job on the `hets` queue fails or exits before marking the `HetsInstance` as `free` again, it needs to be reset. Since the HTTP connection to Hets is closed after 60 seconds, we should force-free the `HetsInstance` after 60 seconds. Mark it as `force-free`.
  * Choosing mechanism (round robin with priorities)
    * If there is a `free` `HetsInstance`, use the first one.
    * If there is no `free` `HetsInstance`, but `force-free` ones exist, choose one of the `force-free` instances by this priority and increase its `queue_size`:
      * smallest `queue_size`
      * smallest `state_updated_at`
    * If there is no `free` and no `force-free` `HetsInstance`, choose a `busy` instance by the same criteria as the `force-free` ones

eugenk:
This shall fix #1445.

This also adds a new rake task `hets: recreate_hets_instances` to create the `HetsIntance`s from the configuration in the database.

To check if really multiple Hets instances are used, I suggest to prove all theorems of an ontology (e.g. v__T) and prove it with several provers. Then, check the `state_updated_at` attributes of the `HetsInstance`s.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -13,6 +13,8 @@ gem 'rdf', '~> 1.1.9'
 gem 'rdf-rdfxml', '~> 1.1.3'
 gem 'rdf-n3', '~> 1.1.2'
 
+gem 'redis-semaphore', '~> 0.2.4'
+
 # Gems used only for assets and not required
 # in production environments by default.
 group :assets do
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -441,6 +441,8 @@ GEM
     redis (3.2.1)
     redis-namespace (1.5.2)
       redis (~> 3.0, >= 3.0.4)
+    redis-semaphore (0.2.4)
+      redis
     ref (1.0.5)
     responders (1.0.0)
       railties (>= 3.2, < 5)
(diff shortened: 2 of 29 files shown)


SGraef:
:+1:

PR #1449, opened 2015-08-10
1448 redesign seeds to be able to use it in production environment

### LINKED ISSUE #1448: redesign seeds to be able to use it in production environment
to be able to use the seeds on test.ontohub.org and develop.ontohub.org we have to change the beaviour of the seeds in production environment.

eugenk:
This shall fix #1448. I checked it on td by running the seeds and there it seems to work (aside from Hets being too old).


SGraef:
:+1:

PR #1456, opened 2015-08-18
Fix proving timeout

eugenk:
This fixes two little issues:

1. Hets closes the connection after 60 seconds. We shouldn't allow timeouts this long. This is supposed to be reverted as soon as we can have an open connection for as long as the job takes.
2. Since the proving pipeline refactoring the timeout is no longer split across the proof obligations. The hint needed to be adjusted accordingly.


tillmo:
:+1:

PR #1462, opened 2015-08-24
Clarify read/write permissions for repositories.

### LINKED ISSUE #1187: clarify read/write permissions for repositories
you can choose between
* private readable - this means not writable at all?
* private readable and writable - this means no public access at all?
* public readable - this means not writable at all, or privately writable, but not publicly?
* public readable and writable

Also, it should be "privately readable" etc.

eugenk:
Shall fix #1187


SGraef:
:+1:

PR #1473, opened 2015-08-28
1450 ontohub slow

SGraef:
This should introduce the rack-mini-profiler gem in development and production
In Production-Mode only an admin can see the miniprofiler and in development it is for every transaction.
this should help to find the weak points in ontohub to make it faster


eugenk:
I can see the profiler badge even when I'm not logged in.

eugenk:
Alright, I see on develop.ontohub.org that it's only visible for admins in production mode.

:+1:

PR #1480, opened 2015-09-09
Fix Semaphore creating a deadlock in synchronous environment.

### LINKED ISSUE #1478: Deadlock in seeds
We encounter a deadlock in the seeds at 030-proof_statuses.

eugenk:
This shall fix #1478.


SGraef:
:+1:

PR #1482, opened 2015-09-28
Fix Semaphore in synchronous environment

eugenk:
For example, parsing failed because Sidekiq::Testing was not defined in
both, the production and development environments.


SGraef:
:+1:

PR #1493, opened 2015-10-13
1490 files in public are not accessible

### LINKED ISSUE #1490: Files in public directory are not accessible
If, for example, we try to access https://ontohub.org/robots.txt, rails doesn't find a route for this URL.
I think this happens because the `*path` route at the end of the routes.rb calls the IRIRouterConstraint which doesn't find an IRI.

SGraef:
should fix #1490

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -701,12 +701,6 @@ is determined according to its path, which is considered as
     as:         :repository_tree,
     constraints: FilesRouterConstraint.new
 
-  get '*path',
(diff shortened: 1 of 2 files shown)


eugenk:
jenkins, test this please

eugenk:
jenkins, test this please

eugenk:
Apart from my previous comment/change:
:+1:

SGraef:
:+1:

PR #1504, opened 2015-10-19
1502 hetsinstance state stays busy after error

### LINKED ISSUE #1502: HetsInstance state stays busy after error
When parsing fails with an error, the selected HetsInstance stays `busy`.

eugenk:
This should fix #1502.


SGraef:
:+1:

PR #1511, opened 2015-10-24
Merge simplecov reports for RSpec and Cucumber.

### LINKED ISSUE #1510: Merge simplecov reports for rspec and cucumber
Currently, simplecov only considers RSpec. It should take Cucumber into account, too.

eugenk:
Should fix #1510.

The `SharedHelper` only existed for simplecov. According to their Readme, one should use the `.simplecov` file instead.

It also adds some coverage groups for the other code parts (fake records, view helpers, serializers).


SGraef:
:+1:

PR #1516, opened 2015-10-29
Fix rake import:logicgraph.

eugenk:
It yielded
```
NoMethodError: undefined method `import_logicgraph' for RakeHelper:Module
```


SGraef:
:+1:

PR #1518, opened 2015-10-30
Add and use queue "hets_load_balancing"

### LINKED ISSUE #1517: Add and use queue hets_load_balancing
The `HetsInstanceForceFreeWorker` seems to enqueue the scheduled jobs instantly. This blocks all other jobs using the same queue. Add a queue only to be used for load balancing done by the `HetsInstanceForceFreeWorker` to allow other jobs to be processed.

eugenk:
Shall fix #1517. I tested it on te and there it seems to work.

### DIFF
--- a/config/god/app.rb
+++ b/config/god/app.rb
@@ -16,6 +16,9 @@ SidekiqWorkers.configure do
       watch 'hets', 1
     end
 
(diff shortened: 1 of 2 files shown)


SGraef:
:+1: looks fine to me

PR #1513, opened 2015-10-26
1497 remove categories for now

SGraef:
should fix #1497

PR #1519, opened 2015-11-02
Delete categories from view in search response.

ebolloff:
We missed something in #1513. The search is broken if there's an ontology with a category.


SGraef:
:+1:

PR #1522, opened 2015-11-06
Perform SInE preprocessing in a transaction

### LINKED ISSUE #1520: Perform SInE preprocessing in a transaction
Perform SInE preprocessing in a transaction. It should render the `cleanup` method obsolete.

eugenk:
Shall fix #1520.

The cleanup is only useful in case of an ungraceful shutdown during the selection. This is the proper use case of a transaction.

I put the whole selection into a transaction to make sure that the finished flag is `true` if and only if the auxiliary objects are available. Thus, the next SineAxiomSelection on the same ontology can reuse them.


tillmo:
:+1:

PR #1524, opened 2015-11-06
1523 hetsinstanceworkers perform too often

### LINKED ISSUE #1523: HetsInstanceWorkers perform too often
A lot of `HetsInstanceForceFreeWorker`s act in a wrong way. Assume the following scenario:

1. A `HetsInstance` is set to `busy`. Then, a `HetsInstanceWorker` is also scheduled to perform the force-freeing in 24 hours.
2. The same `HetsInstance` is set to busy after 23 hours and it doesn't respond for two hours. (Again, a `HetsInstanceWorker` is scheduled).
3. 24 hours after the first step, the first scheduled `HetsInstanceWorker` runs and force-frees the running `HetsInstance` from step two.

`HetsInstanceWorker` should only be allowed to force-free a `HetsInstance`, if the `HetsInstance` was called in the same procedure as the `HetsInstanceWorker` was scheduled. Otherwise, the `HetsInstanceWorker` should do nothing.

eugenk:
This shall fix #1523. Unfortunately, it is not possible to test Sidekiq's **scheduling**.

It seems to work though:
1. Run the seeds
2. Open the http://localhost:3000/admin/sidekiq/scheduled in one tab and http://ontohub.dev/default/partial_order//v__T//antisymmetric///proof_attempts in another tab
3. Create a ProofAttempt for `antisymmetric`.
4. As soon as the proof is finished, check back at the sidekiq/scheduled tab. There should be no additional scheduled jobs.

### DIFF
--- a/app/models/hets_instance.rb
+++ b/app/models/hets_instance.rb
@@ -95,6 +95,7 @@ has a minimal Hets version of #{Hets.minimal_version_string}
 
   def finish_work!
     reload
+    Sidekiq::Status.unschedule(@force_free_job_id)
     self.queue_size -= 1 if queue_size > 0
     if queue_size > 0
       set_busy!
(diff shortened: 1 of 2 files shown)


SGraef:
:+1:

PR #1526, opened 2015-11-06
1525 re check up states of hetsinstances if all are down

### LINKED ISSUE #1525: Re-check up-states of HetsInstances if all are down
When all HetsInstance are down, check their up-states again synchronously.

If all instances are down, the background processing fails. Checking again increases the chances of Ontohub being responsive.

eugenk:
This shall fix #1525.


SGraef:
:+1: looks fine to me

PR #1531, opened 2015-11-09
Reduce scheduled timeout workers

eugenk:
The timeout is only needed to cancel parsing jobs that take longer than
the configured amount of time.
Once the job is done, the TimeoutWorker job can be unscheduled.

This is very similar to #1524 and can't be tested for the same reasons.

This also removes an unused constant.


SGraef:
:+1:

PR #1533, opened 2015-11-10
Fix proving HTTP timeout

eugenk:
The timeout for the prover is not the same as for the HTTP session.

The HTTP session can take much longer than the actual proof because Hets analyses the ontology
again for proving.


tillmo:
:+1:

PR #1534, opened 2015-11-16
Update gems

eugenk:
This updates the bundle again. The latest updates are poltergeist and pg, both fixing some bugs (segfaults).


SGraef:
:+1:

PR #1538, opened 2015-11-16
Remove puma from eye config.

### LINKED ISSUE #1537: Remove puma from eye config
Puma needs to be started as another user than `ontohub`, but eye is managed by `ontohub`.
Remove puma from the eye-config.

### LINKED ISSUE #1645: eye causes ~100% cpu load /puma
On ta it seems, that eye tries to start a puma instance, which fails, because there is already an instance running, which listens on port 3000. So it probably fails and seems to try start it again forever (doesn't give up), which in turn keeps a single CPU ~100% busy ...

eugenk:
Shall fix #1537. 

Edit: And it shall fix #1645.

### DIFF
--- a/config/processes/app.eye
+++ b/config/processes/app.eye
@@ -40,30 +40,4 @@ Eye.application :ontohub do
       end
     end
   end
-
-  process :puma do
-    ctrl_socket = 'unix:///tmp/pumactl.sock'
-    home = Rails.root.join('..').to_s
-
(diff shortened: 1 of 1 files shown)


eugenk:
We don't want this anymore.

eugenk:
I must correct: We want to have this.

ebolloff:
:+1:

PR #1556, opened 2016-01-07
167 creating a new ontology version needs to be propagated

### LINKED ISSUE #167: creating a new ontology version needs to be propagated
... along all import links. That is, if O2 imports O1, and O1 is changed to a new version, the entities and sentences of O2 need to be recomputed, and a new version for O2 needs to be created.

eugenk:
This shall fix #167.

All ontologies T that have a mapping from S are reanalysed after S is updated if S and T are in different files and the file of T is not changed in the same commit.

This pull request moves some code to new files for a clearer structure. I recommend to read this commit by commit.


SGraef:
i hope we don't blow up our load to much but the code looks okay for me :+1:

eugenk:
Rebase conflicts were trivial.
Caused by: Move SequentialWorker to a separate file. adcdabf

I'm merging this as soon as the CI results are green.

PR #1558, opened 2016-01-07
1383 cloning of private repositories is possible

### LINKED ISSUE #1383: Cloning of private repositories is possible
It is possible to clone a private repository via the git daemon. This should be restricted. I suggest, we only allow to clone private repositories via git-ssh.

eugenk:
This shall fix #1383 and is deployed on develop. It renames one and adds another settings key. When deploying on another machine, one needs to think about it.


eugenk:
Put the new settings keys into the deployment wiki

eugenk:
Oh, I see that I already did that in 6bbb71b.

SGraef:
:+1:

PR #1570, opened 2016-01-15
Reactivate Hound

eugenk:
This updates the config file for hound to be compliant with the current rubocop version.


SGraef:
:+1:

PR #1573, opened 2016-01-17
Add hound configs

eugenk:
This adds default and activates config files for the different linters that Hound uses.

The only change to the default configs that I made is to remove the limit of the line length for HAML.


SGraef:
:+1:

PR #1576, opened 2016-01-23
Fix proof statuses

eugenk:
This solves two issues with the SZS ontology:
* Prefixes had a double `http://`
* Label was ambiguous due becase a wrong prefix was used


SGraef:
:+1:

PR #1580, opened 2016-01-27
1579 HetsInstance can't set up state to false

### LINKED ISSUE #1579: HetsInstance can't set up state to false
When checking the up state (e.g. on each `save`), the `VersionCaller` checks if the instance of hets is reachable. If, however, the host is down or does not respond in time, this raises an error and the `up` attribute is never set to `false`.

eugenk:
This shall fix #1579


SGraef:
:+1:

PR #1595, opened 2016-02-08
635 backup

eugenk:
This PR
* changes the target directory for backups
* creates it if it doesn't exist
* fails if this target path is a file and not a directory
* uses exit error codes

for the backup script.


SGraef:
:+1:

PR #1598, opened 2016-02-17
Sleep instead of waiting actively.

eugenk:
One CPU core was always at 100% because of waiting actively for an interrupt. With this change, the process is simply asleep forever.


SGraef:
:+1:

PR #1599, opened 2016-02-18
Add hets-server development compatibility.

eugenk:
This allows invoker to start `hets-server` if it is installed instead of `hets`


SGraef:
:+1:

PR #1600, opened 2016-02-18
1184 wrong class hierachy tree is displayed

### LINKED ISSUE #1184: Wrong class hierachy tree is displayed
see http://ontohub.org/sandbox/Family///symbols?kind=Class#detail
When clicking on "Class Details", 7 classes are shown, but the hierarchy tree shows only 3 classes.

SGraef:
should fix #1184

### DIFF
--- a/app/controllers/symbols_controller.rb
+++ b/app/controllers/symbols_controller.rb
@@ -18,9 +18,13 @@ class SymbolsController < InheritedResources::Base
     if ontology && (ontology.is?('OWL2') || ontology.is?('OWL'))
       begin
         symbols = ontology.symbol_groups
-        extra_notes = ontology.symbol_groups.where('id NOT IN (SELECT DISTINCT(parent_id) FROM e_edges UNION SELECT DISTINCT(child_id) FROM e_edges)')
+        request = <<-SQL
+          id NOT IN (SELECT DISTINCT(parent_id) FROM e_edges
+           UNION SELECT DISTINCT(child_id) FROM e_edges)
+        SQL
+        extra_notes = ontology.symbol_groups.where(request)
         @nodes = SymbolGroup.roots_of(*symbols)
-        @nodes = @nodes + extra_notes
+        @nodes += extra_notes
       rescue
         @nodes = []
       end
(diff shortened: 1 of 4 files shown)


lib/tarjan_tree.rb
```diff
@@ -1,24 +1,27 @@
 # Class for using the Tarjan algorithm to remove cycles in the symbol trees.
 class TarjanTree
   include TSort
-  attr_accessor :hashed_symbols, :subclasses, :ontology
+  attr_accessor :hashed_symbols, :subclasses, :ontology, :symbols, :subclass_sym
```
eugenk: Please use a plural named variable like `subclass_symbols` instead of `subclass_sym`.

lib/tarjan_tree.rb
```diff
@@ -52,10 +55,15 @@ def create_tree(ontology)
   end
 
   def create_groups(ontology)
+    symbols = self.symbols - subclass_sym
     strongly_connected_components.each do |symbol_ids|
-      symbols = OntologyMember::Symbol.find(symbol_ids)
-      name = group_name_for(symbols)
-      SymbolGroup.create!(ontology: ontology, symbols: symbols, name: name)
+      syms = OntologyMember::Symbol.find(symbol_ids)
```
eugenk: Please give a more suitable name to the variable `syms`. How about `strongly_connected_symbols` or `equivalent_symbols` or `something_else_symbols`?

app/controllers/symbols_controller.rb
```diff
@@ -18,7 +18,13 @@ def index
     if ontology && (ontology.is?('OWL2') || ontology.is?('OWL'))
       begin
         symbols = ontology.symbol_groups
+        request = <<-SQL
+          id NOT IN (SELECT DISTINCT(parent_id) FROM e_edges
+           UNION SELECT DISTINCT(child_id) FROM e_edges)
+        SQL
+        extra_notes = ontology.symbol_groups.where(request)
```
eugenk: I suppose you meant `extra_nodes` with a "d" ;)
In what way are these "extra"? It seems like they are in a subclass hierarchy somewhere between the child and the parent, but I'm not sure if my guess is correct.

eugenk:
Please add a test

eugenk:
:+1:

PR #1601, opened 2016-02-19
Fix symbol parsing and code style

eugenk:
This fixes three little issues:
* **During** parsing, `ontology.symbols` was empty, even if there are symbols. This made parsing callbacks that depend on this association fail.
* The method `update_or_create_from_hash` is supposed to return the symbol, but it returned a boolean.
* Code style: Indentation and the place of the association definition were wrong.

I add no test because this only affects the parsing code. I would need to intercept the parsing and then test internals (which is bad style). The effect will be tested implicitly by #1600.


SGraef:
:+1:

Issue #1450, opened 2015-08-13
Ontohub  slow


fabianneuhaus:
When I try to browse my ontologies, the server is very slow. Admittedly, Ontohub isn't very fast at the best of times, but we are talking 30 seconds and more for opening a page, e.g.,while I am writing this bug report I am  waiting for
https://ontohub.org/repositories/imageschemafamily

fabianneuhaus:
Bad Gateway!

The proxy server received an invalid response from an upstream server.

The proxy server could not handle the request GET /.

Reason: Error reading from remote server

If you think this is a server error, please contact the webmaster.

Error 502

ta.iws.cs.ovgu.de
Apache/2.4.10 (Ubuntu) OpenSSL/1.0.1f

tillmo:
@jelmd, could you please have a look at this?

jelmd:
Probably a temp. error.
Might correlate to ... ?
```
2015-07-21 21:25:10 UTC [4039-1] ontohub@ontohub LOG:  sending cancel to blocking autovacuum PID 4246
2015-07-21 21:25:10 UTC [4039-2] ontohub@ontohub DETAIL:  Process 4039 waits for AccessExclusiveLock on relation 20079 of database 16385.
2015-07-21 21:25:10 UTC [4039-3] ontohub@ontohub STATEMENT:  ALTER TABLE "sentences" ADD COLUMN "state" character varying(255)
2015-07-21 21:25:10 UTC [4246-1] ERROR:  canceling autovacuum task
2015-07-21 21:25:10 UTC [4246-2] CONTEXT:  automatic vacuum of table "ontohub.public.sentences"
```

The other odd thing I saw is, that there were 9 pretty busy psql processes causing a load of ~5. Not that dramatic wrt. resources, but odd, since I guess, the DB is not yet really challenged by ontohub. I guess it is more busy with maintaining itself than with "external" tasks, so perhaps psql needs some tuning. For now I've rosen  checkpoint_segments from 3 to 7 (suggested by psql logs) ...

tillmo:
@jelmd: this is from July 21, and probably related to the move of the Ontohub server. In regular operation mode, there are no `ALTER TABLE`s.

tillmo:
@SGraef will have a look at https://github.com/MiniProfiler/rack-mini-profiler and https://github.com/ruby-prof/ruby-prof. If these do not help, we will use http://newrelic.com/ruby/rails

fabianneuhaus:
[...]

PR #1618, opened 2016-03-10
Improve locId

### LINKED ISSUE #1630: RoutingError after destroying and recreating
When I create a Repository with name "demo"
And I upload [periods.dol](https://ontohub.org/repositories/colore-sample/d290f729ab4a630d212a9114c88fdbf24ea689d3/show/periods/mappings/periods.dol)
And I attempt to prove every Theorem in its child ontology "i2__T"
And I destroy the repository via `Repository.find_by_path('demo').destroy`
And I create a Repository with name "demo"
And I upload [periods.dol](https://ontohub.org/repositories/colore-sample/d290f729ab4a630d212a9114c88fdbf24ea689d3/show/periods/mappings/periods.dol)
And I attempt to prove every Theorem in its child ontology "i2__T"
And I visit http://localhost:3000/demo/periods//i2__T//Ax_0_1//proof-attempt-1
Then I get a RoutingError:
```
No route matches [GET] "/demo/periods/i2__T/Ax_0_1/proof-attempt-1"
```

SGraef:
This PullRequest should help to make Ontohub a little bit faster. The LocID as attribute of the objects results in a lot of queries. So we decided to introduce an new model.

For review purpose the pull request  is made before new tests are implemented

- [x] fix tests
- [x] add new tests
- [x] catch PG::UniqueViolation error to restrict namespace of sentences/symbols.
- [x] fix routing errors
- [x] add migration to give every symbol a locid that does not have one
- [ ] set minimum Hets version to what https://github.com/spechub/Hets/pull/1630 results in.

This shall also fix #1630.

--

*Don't merge this before https://github.com/spechub/Hets/issues/1565 is done.*


eugenk:
Note: This pull request shall resolve #1450

app/models/loc_id_base_model.rb
```diff
@@ -0,0 +1,36 @@
...
+ def destroy_locid
+     LocId.where(assorted_object_id: self.id, assorted_object_type: self.class).first.destroy
+ end
+
+ def locid
+   LocId.where(assorted_object_id: self.id, assorted_object_type: self.class).first.try(:locid)
+ end
+
+ def locid=(string)
+   locid = LocId.where(assorted_object_id: self.id, assorted_object_type: self.class).first
```
eugenk: Please put `LocId.where(assorted_object_id: self.id, assorted_object_type: self.class).first` into a private method `query_locid` and call it instead. You use it three times in the lines above.

app/models/ontology/class_methods_and_scopes.rb
```diff
@@ -76,7 +76,7 @@ module Ontology::ClassMethodsAndScopes
 
   module ClassMethods
     def find_with_locid(locid, iri = nil)
```
eugenk: Isn't this exactly what you do in the parent class? This method seems redundant.

app/models/loc_id_base_model.rb
eugenk: I mean protected, not private. Use `query_locid` in the subclasses as well.

app/models/loc_id_base_model.rb
```diff
@@ -0,0 +1,36 @@
+class LocIdBaseModel < ActiveRecord::Base
+ @abstract_class = true
+ has_many :loc_ids, as: :assorted_object
```
eugenk: Why does it have many and not one?

app/models/ontology/class_methods_and_scopes.rb
```diff
@@ -88,7 +88,9 @@ def find_with_locid(locid, iri = nil)
 
     def find_with_iri(iri)
       locid = iri_to_locid(iri)
-      ontology = where('locid LIKE ?', '%' << locid).first
+      ontology = ::LocId.where(
+                               'locid LIKE ?', '%' << locid
+                              ).first.try(:assorted_object)
```
eugenk: Please use the agreed code style:
```ruby
method_call(arg1,
            arg2, 
            arg3, 
            arg4)
```
having the opening parenthesis on the line of the first argument and the closing one on the line of the last argument.

app/models/ontology/symbols.rb
```diff
@@ -34,16 +34,10 @@ def update_or_create_from_hash(hash, timestamp = Time.now)
         sep = '//'
         locid_portion =
           if e.name.include?('://')
-            if e.label
-              e.label
-            else
-              portion = e.name.end_with?('>') ? e.name[0..-2] : e.name
-              portion = portion.split('#', 2).last
-            end
+            Rack::Utils.escape_path(e.name)
```
eugenk: Why don't you use the label any more?

app/models/ontology_member/symbol.rb
```diff
@@ -16,13 +16,14 @@ class Symbol < ActiveRecord::Base
                                      dependent: :destroy
     has_many :sine_symbol_axiom_triggers, dependent: :destroy
 
-    attr_accessible :locid
+    has_many :loc_ids, as: :assorted_object
```
eugenk: Why many instead of one?

app/models/loc_id_base_model.rb
eugenk: As I think about it, private seems to be the better choice because you don't need exactly this in the subclasses.

app/models/proof_status.rb
```diff
@@ -1,4 +1,4 @@
-class ProofStatus < ActiveRecord::Base
+class ProofStatus < LocIdBaseModel
```
eugenk: IRRC this class should switch from LocId to slug based routing in this pull request ;)

eugenk: But we can change this in a future pull request as well.

db/migrate/20160310132620_move_loc_id_to_own_model.rb
```diff
@@ -0,0 +1,53 @@
+class MoveLocIdToOwnModel < MigrationWithData
+  def up
```
eugenk: What do we do with duplicate legacy locids? Re-analyze the ontology? @tillmo: What do you think?

lib/ontology_saver.rb
```diff
@@ -104,16 +104,16 @@ def find_existing_ontology(ontology_version_options)
       without_parent.first
   end
 
-  def create_ontology(filepath, locid)
+  def create_ontology(filepath, basepath)
```
eugenk: argument `basepath` is redundant. it is set in `ontology.basepath = File.basepath(filepath)`

spec/factories/ontology_factory.rb
```diff
@@ -17,13 +17,18 @@
...
+        locid = "#{ontology.parent.locid}//#{ontology.name}"
+      else
+        locid = "/#{ontology.repository.path}/#{ontology.name}"
+      end
+       LocId.where(
+                    locid: locid
+                  ).first_or_create!(
+                  assorted_object_id: ontology.id,
+                  assorted_object_type: 'SingleOntology',
+                  )
```
eugenk: Please use the agreed upon code style:
```ruby
       LocId.where(locid: locid).
         first_or_create!(assorted_object_id: ontology.id,
                          assorted_object_type: 'SingleOntology')

```

spec/factories/mapping_factory.rb
```diff
@@ -9,8 +9,13 @@
...
     association :ontology
 
-    after(:build) do |mapping|
-      mapping.locid = "#{mapping.ontology.locid}//#{mapping.name}"
+    after(:create) do |mapping|
+      LocId.where(
+                    locid: "#{mapping.ontology.locid}//#{mapping.name}",
+                  ).first_or_create!(
+                  assorted_object_id: mapping.id,
+                  assorted_object_type: 'Mapping',
```
eugenk: This may not be a relevant request, but I'd rather use `Mapping` instead of `'Mapping'` because then we should get a better error if there is something wrong with class loading or typos.

spec/factories/proof_status_factory.rb
```diff
@@ -78,8 +78,16 @@
 
     initialize_with do
       statuses.map do |status|
-        ProofStatus.where(identifier: status['identifier']).any? ||
-          ProofStatus.create(status)
+        if ProofStatus.where(identifier: status['identifier']).any?
+          pr_status = ProofStatus.where(identifier: status['identifier']).first
+        else
+          pr_status = ProofStatus.create(status)
+        end
```
eugenk: Suggestion:
```ruby
proof_status = ProofStatus.where(identifier: status['identifier']).first
proof_status ||= ProofStatus.create(status)
```

app/models/loc_id_base_model.rb
```diff
@@ -0,0 +1,36 @@
...
+   if table_name == "ontologies"
+     if result.nil? && iri
+       result = AlternativeIri.where('iri LIKE ?', '%' << iri).
+        first.try(:ontology)
+     end
+   end
+   result
+ end
+
+ def destroy_locid
```
eugenk: Please make this method protected.

db/migrate/20160310132620_move_loc_id_to_own_model.rb
tillmo: yes, re-analyse. Is it guaranteed that this will eliminate the duplication?

eugenk: As soon as Hets implements that symbols and sentences are not allowed to have the same name, the corresponding version will fail and the Ontohub-database will not have the problem of duplicates.

spec/factories/mapping_factory.rb
tillmo: It is relevant, I strongly agree!

app/models/loc_id_base_model.rb
```diff
@@ -0,0 +1,36 @@
...
+   LocId.where(assorted_object_id: self.id, assorted_object_type: self.class).first.try(:locid)
+ end
+
+ def locid=(string)
+   locid = LocId.where(assorted_object_id: self.id, assorted_object_type: self.class).first
+   if locid
+     locid.update_attributes(locid: string)
+   else
+     LocId.create(assorted_object_id: self.id,
+                  assorted_object_type: self.class,
```
eugenk: Please use `.class.to_s` instead of `.class` everywhere.

app/models/loc_id.rb
```diff
@@ -0,0 +1,5 @@
+class LocId < ActiveRecord::Base
+  belongs_to :assorted_object, polymorphic: true
+  attr_accessible :assorted_object, :assorted_object_id, :assorted_object_type
```
eugenk: I suggest to rename this to `attr_accessible :object, :object_id, :object_type `

spec/factories/symbol_factory.rb
```diff
@@ -21,13 +21,27 @@
...
     name { FactoryGirl.generate :name }
 
-    after(:build) do |symbol|
-      symbol.locid = "#{symbol.ontology.locid}//#{symbol.name}"
+    after(:create) do |symbol|
+      LocId.where(
+                    locid: "#{symbol.ontology.locid}//#{symbol.name}",
+                  ).first_or_create!(
+                  assorted_object_id: symbol.id,
+                  assorted_object_type: 'Sentence',
```
eugenk: This should not be `'Sentence'` because it's in the *symbol* factory.
I fixed this in 5a76ecf3e9542845fe05f65e722394fd5228a85a.

spec/factories/ontology_factory.rb
```diff
@@ -17,13 +17,18 @@
...
+        locid = "#{ontology.parent.locid}//#{ontology.name}"
+      else
+        locid = "/#{ontology.repository.path}/#{ontology.name}"
+      end
+       LocId.where(
+                    locid: locid
+                  ).first_or_create!(
+                  assorted_object_id: ontology.id,
+                  assorted_object_type: ontology.class.to_s,
+                  )
```
eugenk: Please use the agreed upon code style:
```ruby
       LocId.where(locid: locid).
         first_or_create!(assorted_object_id: ontology.id,
                          assorted_object_type: ontology.class.to_s)
```

app/models/ontology_member/symbol.rb
```diff
@@ -35,5 +36,11 @@ def self.groups_by_kind
     def to_s
       display_name || name
     end
+
+    def locid
```
eugenk: Redundant method (this is defined in the parent class)

eugenk:
Now this branch is **based on the branch of #1649 (1509-switch_proofstatus_from_loc_id_to_slug) instead of staging**.

eugenk:
Don't merge this before https://github.com/spechub/Hets/issues/1565 is done.

db/migrate/20160310132620_move_loc_id_to_own_model.rb
eugenk: Done with 6a3d6da.

app/models/ontology/class_methods_and_scopes.rb
eugenk: Removed with bceb848.

app/models/ontology_member/symbol.rb
eugenk: This is useless code anyway. Removed with fc7dd4a.

lib/ontology_saver.rb
eugenk: Fixed with bceb848.

app/models/loc_id.rb
eugenk: Done with 682aba7.

app/models/ontology_member/symbol.rb
eugenk: Fixed with bceb848.

lib/ontology_saver.rb
```diff
@@ -104,12 +104,10 @@ def find_existing_ontology(ontology_version_options)
...
```
eugenk: Fixed with 57854d6.

app/models/ontology.rb
```diff
@@ -1,4 +1,4 @@
-class Ontology < ActiveRecord::Base
+class Ontology < LocIdBaseModel
```
eugenk: I don't want to split it into more Concerns now.

eugenk: I don't want to split it into more Concerns now.

db/migrate/20160310132620_move_loc_id_to_own_model.rb
```diff
@@ -0,0 +1,57 @@
+require_relative '../data/20160310132620_move_loc_id_to_own_model_data.rb'
+class MoveLocIdToOwnModel < MigrationWithData
+  CLASSES = [Mapping,
```
eugenk: Done with a5d2be1.

db/data/20160310132620_move_loc_id_to_own_model_data.rb
```diff
@@ -0,0 +1,36 @@
...
+class MoveLocIdToOwnModelData < ActiveRecord::Migration
+  # do nothing
+  def self.up
+  end
+
+  # do nothing
+  def self.down
+  end
+
+  def self.reanalyze_duplicates(duplicate_locid_objects)
```
eugenk: Fixed with e407c7f.

app/models/ontology.rb
eugenk: I don't want to split it into more Concerns now.

db/data/20160310132620_move_loc_id_to_own_model_data.rb
```diff
@@ -0,0 +1,37 @@
...
```
eugenk: Fixed with e407c7f.

db/data/20160310132620_move_loc_id_to_own_model_data.rb
```diff
@@ -0,0 +1,40 @@
...
+    end
+
+    ontologies_to_parse.uniq.each do |ontology|
+      # The ontologies need to be parsed asynchronously because the HTTP server
+      # does not respond during the migration.
+      ontology.current_version.async_parse
+    end
+  end
+
+  def self.ontology_to_parse(duplicate)
```
eugenk: I think this is okay.

eugenk:
http://localhost:3000/default/Subclass//SubclassToleranceOne///theorems fails with:
undefined method `gsub' for nil:NilClass
app/controllers/application_controller.rb:82 (locid_for)

eugenk:
> http://localhost:3000/default/Subclass//SubclassToleranceOne///theorems fails with:
undefined method `gsub' for nil:NilClass
app/controllers/application_controller.rb:82 (locid_for)

This seems to have resolved itself...

lib/router_constraints.rb
```diff
@@ -58,28 +62,54 @@ def matches?(request, path = nil)
...
+    self.class.instance_variable_set(:@elements_cache_clear_at, next_clearing)
+
+    self.class.instance_variable_set(:@elements_cache, {})
+  end
+
+  def elements_cache_clearing_scheduled?
+    Time.now > self.class.instance_variable_get(:@elements_cache_clear_at)
+  end
+
+  def assign_path_parameters(request, element)
```
eugenk: I can't see a simpler way to implement this. If anyone has an idea, I'm glad to change it. Otherwise let's keep it as is.

db/seeds/040-git.rb
```diff
@@ -31,7 +32,10 @@
   basename = File.basename(path)
 
   version = repository.save_file path, basename, "#{basename} added", @user, do_not_parse: true
-  version.parse
+  begin
+    version.parse
+  rescue Hets::SyntaxError
```
eugenk: But this is exactly what I want to do in the seeds. See the comment below.

ebolloff:
@phyrog could you take a look at this, too?

eugenk:
jenkins, test this please

eugenk:
As soon as both of you have reviewed this pull request the first time, I would like to merge it to develop. Then, we should make extensive testing on develop.ontohub.org before you sign this off.

eugenk:
The routing errors seem to be a bug in the loading of the Rails application:.

Given I have a breakpoint at `lib/router_constraints.rb:69`
When I start invoker
And I initialize the database and run the seeds
And I visit an ontology
Then I get a RoutingError
And I get in the debugging console:
```ruby
[9] pry(#<LocIdRouterConstraint>)> element.class.to_s
=> "SingleOntology"
[10] pry(#<LocIdRouterConstraint>)> @find_in_klass.to_s
=> "Ontology"
[11] pry(#<LocIdRouterConstraint>)> element.is_a?(@find_in_klass)
=> false
```

When I additionally stop and restart the rails server with
```text
invoker remove ontohub
invoker add ontohub
```
And I visit the ontology again
Then I don't get a RoutingError
And I get in the debugging console:
```ruby
[2] pry(#<LocIdRouterConstraint>)> element.class.to_s
=> "SingleOntology"
[3] pry(#<LocIdRouterConstraint>)> @find_in_klass.to_s
=> "Ontology"
[4] pry(#<LocIdRouterConstraint>)> element.is_a?(@find_in_klass)
=> true 
```
I tested it with puma in development mode (instead of webrick) and I get the same behavior.

eugenk:
Maybe it's not a bug, but a behavior we should address when deploying. Every time we execute migrations, we need to restart the web server afterwards. I don't know if this is done already, but we need to keep it in mind.

lib/router_constraints.rb
```diff
@@ -58,28 +62,53 @@ def matches?(request, path = nil)
...
+    ontology = element.is_a?(Ontology) ? element : element.ontology
+
+    proof_attempt = element.proof_attempt if @map[:proof_attempt]
+    theorem = element.theorem if @map[:theorem]
```
eugenk: Because of the eval a few lines down, this is actually used.

lib/router_constraints.rb
```diff
@@ -58,28 +62,53 @@ def matches?(request, path = nil)
...
+  end
+
+  def elements_cache_clearing_scheduled?
+    Time.now > self.class.instance_variable_get(:@elements_cache_clear_at)
+  end
+
+  def assign_path_parameters(request, element)
+    ontology = element.is_a?(Ontology) ? element : element.ontology
+
+    proof_attempt = element.proof_attempt if @map[:proof_attempt]
```
eugenk: Because of the eval a few lines down, this is actually used.

lib/router_constraints.rb
```diff
@@ -58,28 +62,53 @@ def matches?(request, path = nil)
...
+    theorem = element.theorem if @map[:theorem]
+
+    path_params = {repository_id: ontology.repository.to_param}
+    %i(proof_attempt theorem ontology element).each do |param|
+      path_params[@map[param]] = eval(param.to_s).id if @map[param]
```
eugenk: It isn't in this case because we only use meta programming to access the variable. I'm not sure, though, if this code gives us any benefit (less readable, only a little more compact, a lot more DRY). @ebolloff and @phyrog, please comment on this.

phyrog: How about `send(param.to_sym)` instead?

phyrog: Looks OK to me.

tillmo:
spechub/Hets#1630 has been merged now.

ebolloff:
Let's try this on develop :+1:

eugenk:
This is now deployed on develop.ontohub.org

db/migrate/20160310132620_move_loc_id_to_own_model.rb
```diff
@@ -0,0 +1,58 @@
...
+             OntologyMember::Symbol,
+             Ontology,
+             ProofAttempt,
+             ProverOutput,
+             Theorem,
+             Axiom].freeze
+  TABLES = CLASSES.map(&:table_name).uniq.freeze
+
+  def up
```
eugenk: I think we can leave this as is. What do you think, @ebolloff, @phyrog?

ebolloff:
https://develop.ontohub.org/mappings

```
 NoMethodError in Mappings#index

Showing /local/home/ontohub/webapp/app/views/mappings/_mapping.html.haml where line #3 raised:

undefined method `gsub' for nil:NilClass
```

eugenk:
>  NoMethodError in Mappings#index

This is fixed now. See the new migration (and the execution order of all new migrations of this branch).

phyrog:
OK, I pretty much clicked every button and link I could find at least 10 times now. While I found some new issues, none were caused by this PR. Just to be sure, I will do another click-session tomorrow, but after that (if there are no new errors) you'll get a :+1: from me.

db/migrate/20160310132620_move_loc_id_to_own_model.rb
phyrog: Yes, I believe for a migration this is acceptable.

eugenk:
The commits are all "new" because I rebased on staging. There were no conflicts.

@phyrog: is this a real sign off or still pending?

PR #1621, opened 2016-03-14
Quick fix #1593.

### LINKED ISSUE #1593: undefined method `retrieve_available_provers' for nil:NilClass
http://ontohub.org/sandbox/engine2///ontology_versions
I have deleted a child ontology, maybe this caused the problem.

eugenk:
This shall fix #1593 in a quick and dirty way to prevent HTTP 5xx errors on ontohub.org.

This is going to be reverted as soon as a proper fix is implemented (which is very complicated).


ebolloff:
:+1:

PR #1624, opened 2016-03-14
1591 child ontologies are indefinitely pending

### LINKED ISSUE #1591: child ontologies are indefinitely pending
If I upload a legal ontology library (=distributed ontology) and then push a new version containing a parsing error, the ontology library is `failed`, but the children are all indefinitely `pending`. See e.g. http://ontohub.org/sandbox/engine

eugenk:
This shall fix #1591.

Most of the changed lines are just fixing test behaviour. Before, we did not just stub the Hets-API call, but the whole parsing process. The last commit fixes this issue by only stubbing Hets and still running the whole rest of the parsing pipeline.


ebolloff:
:+1:

PR #1628, opened 2016-03-15
Change delegates to associations.

### LINKED ISSUE #1625: ProverOutput#ontology delegated to proof_attempt.ontology, but proof_attempt is nil
We get
```
ProverOutput#ontology delegated to proof_attempt.ontology, but proof_attempt is nil
```
on http://tc.iws.cs.ovgu.de/ontohub-demo/periods//i2__T//Ax_0_1//proof-attempt-1//prover-output

Make the delegation an association with `through` to fix this.

eugenk:
This shall fix #1625.


ebolloff:
:+1:

PR #1638, opened 2016-03-23
1637 display information about mappings in the graphs view

### LINKED ISSUE #1637: Display information about mappings in the Graphs view
When I visit the Graphs tab of an ontology with Mappings
And I click on an edge representing a Mapping
Then I should see some information about the Mapping on the right side of the page.

But I don't...

eugenk:
This shall fix #1637.


app/controllers/graphs_controller.rb
```diff
@@ -32,6 +32,22 @@ def graph_data
...
+    items.map { |item| populate_with(item, *attributes) }
+  end
+
+  def populate_with(original_item, *attributes)
+    item = original_item.as_json.dup
+    attributes.each do |attribute|
+      if original_item.respond_to?(attribute) && !item.key?(attribute.to_s)
+        item = item.merge(attribute.to_s => original_item.send(attribute))
+      else
+        item
```
ebolloff: I think line 44-45 are redundant.

eugenk: Fixed with 9bc38e6e1a21fc5634d3c4ed2cbeafeeac66a174

ebolloff:
:+1:

PR #1650, opened 2016-04-16
1633 add edit and show buttons to ontology versions

### LINKED ISSUE #1633: Add edit button to versions list
For failed ontologies like version 1 in http://ontohub.org/sandbox/syntax%20error/periods///ontology_versions, we want to have an edit button under the red error message box.

This edit button must only be shown if the failed version is the most recent version and if the user is allowed to edit the file. You can see the permission check at the edit button on file view (see the following link).

Clicking this button shall switch the page to the [file view](http://ontohub.org/sandbox/syntax%20error/periods.dol) and act as if the "Edit file" button was clicked on page load.

You can see the corresponding coffeescript of the file view/edit function in [app/assets/javascripts/codemirror-custom.js.coffee](https://github.com/ontohub/ontohub/blob/3aa9e3f744713c1a073dd004fbb77b04ac0b6e0c/app/assets/javascripts/codemirror-custom.js.coffee)
and the edit button in the file view in [app/views/files/show.html.haml:26](https://github.com/ontohub/ontohub/blob/3aa9e3f744713c1a073dd004fbb77b04ac0b6e0c/app/views/files/show.html.haml#L26)

phyrog:
This fixes #1633. Adds edit and show buttons to ontology versions depending on the following factors:

- failed latest version when having write access: edit button
- else: view button to file version at that point

### DIFF
--- a/app/helpers/ontology_version_helper.rb
+++ b/app/helpers/ontology_version_helper.rb
@@ -1,0 +1,29 @@
+module OntologyVersionHelper
+  def failed?(ontology_version)
+    ontology_version.state == 'failed'
+  end
+
+  def can_edit?(ontology_version)
+    can?(:write, ontology_version.repository) &&
+      failed?(ontology_version) &&
+      ontology_version.latest_version?
+  end
+
+  def btn_url(ontology_version)
+    fancy_repository_path(resource_chain[0],
+                          path: ontology_version.path,
+                          ref: ontology_version.commit_oid,
+                          action: :show,
+                          exact_commit: !ontology_version.latest_version?)
+  end
+
+  def edit_or_view_button(ontology_version)
+    editable = can_edit?(ontology_version)
+    btn_text = editable ? 'Edit' : 'View'
+    anchor = editable ? '#edit' : ''
+
+    link_to btn_text,
+            btn_url(ontology_version) + anchor,
+            class: 'btn btn-xs btn-default'
+  end
+end
--- a/app/views/ontology_versions/_ontology_version.html.haml
+++ b/app/views/ontology_versions/_ontology_version.html.haml
@@ -4,4 +4,6 @@
   %td= timestamp ontology_version
   %td= ontology_version.user ? fancy_link(ontology_version.user) : '-'
   %td= render partial: 'shared/state_tag', locals: {resource: ontology_version}
-  %td= link_to 'Download', fancy_repository_path(resource_chain[0], path: ontology_version.path, ref: ontology_version.commit_oid, action: :download), class: 'btn btn-xs btn-default'
+  %td
+    = link_to 'Download', fancy_repository_path(resource_chain[0], path: ontology_version.path, ref: ontology_version.commit_oid, action: :download), class: 'btn btn-xs btn-default'
+    = edit_or_view_button(ontology_version)
--- a/features/OntologyVersions.feature
+++ b/features/OntologyVersions.feature
@@ -18,7 +18,7 @@ Scenario: Displaying an edit button on the latest ontology if it is failed with 
   And I am logged in
   And there is an ontology with a "failed" version
   And I have permissions to edit the ontology
-  And there is a ontology file
+  And there is an ontology file
   When I visit the versions tab of the ontology
   And I should see a "Edit" button for the latest version
   And I should see a "View" button for every other version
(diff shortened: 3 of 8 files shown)


phyrog:
I will add some cucumber tests for this on monday

app/views/ontology_versions/_ontology_version.html.haml
```diff
@@ -4,4 +4,14 @@
...
+    = link_to 'Download', fancy_repository_path(resource_chain[0], path: ontology_version.path, ref: ontology_version.commit_oid, action: :download), class: 'btn btn-xs btn-default'
+    - latest_version = ontology_version_counter == 0
+    - if can?(:write, ontology.repository) && ontology_version.state == 'failed' && latest_version
+      - btn_text = 'Edit'
+      - anchor = '#edit'
+    - else
+      - btn_text = 'View'
+      - anchor = ''
+    - exact_commit = !latest_version
+    = link_to btn_text, fancy_repository_path(resource_chain[0], path: ontology_version.path, ref: ontology_version.commit_oid, action: :show, exact_commit: exact_commit) + anchor, class: 'btn btn-xs btn-default'
```
eugenk: Please put the business logic (assigning the variables) of the view into a helper method to keep the view slim and stupid. You could, for instance, put all the code in the `td` except for the download button into a method `edit_or_view_button`.

phyrog: Alright, done. Had to split it up a lot to get it past hound, though.

app/helpers/ontology_version_helper.rb
```diff
@@ -0,0 +1,32 @@
+module OntologyVersionHelper
+  def failed?(ontology_version)
+    ontology_version.state == 'failed'
+  end
+
+  def latest_version?(ontology_version)
+    ontology_version.ontology.versions.last == ontology_version
```
phyrog: I'm not sure if it's better to check it this way (was using the iteration variable from the view before). Probably more robust, but definitely slower.

eugenk: I agree that this is the most robust version and we should stick with it.

Also, I think we could use this as a method of the model `OntologyVersion`.

app/helpers/ontology_version_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+  end
+
+  def can_edit?(ontology, ontology_version)
+    can?(:write, ontology.repository) &&
```
eugenk: You don't need the argument `ontology` here. There is an association `repository` on `OntologyVersion`.

app/helpers/ontology_version_helper.rb
```diff
@@ -0,0 +1,32 @@
...
+                          path: ontology_version.path,
+                          ref: ontology_version.commit_oid,
+                          action: :show,
+                          exact_commit: !latest_version?(ontology_version))
+  end
+
+  def edit_or_view_button(ontology, ontology_version)
+    btn_text = can_edit?(ontology, ontology_version) ? 'Edit' : 'View'
+    anchor = can_edit?(ontology, ontology_version) ? '#edit' : ''
+
```
eugenk: Please cache the result of `can_edit` in a variable instead of calling it twice.

phyrog:
Ok, I wrote some tests for the buttons, but somehow after opening the edit button, the File browser is shown (instead of the view/edit page). `current_path` shows the correct path, but `save_and_open_page` definitely shows the wrong page.

eugenk:
I think that's because there is no file for the ontology in the git repository. Our factories skip the file creation because we don't need them most of the time. You can add a file for the ontology with something like
```ruby
  repository = @ontology.repository
  user = repository.permissions.first.subject
  Tempfile.create('testfile') do |f|
    repository.save_file_only(f.path, @ontology.path, 'add ontology file', user)
  end
```

features/OntologyVersions.feature
```diff
@@ -18,6 +18,7 @@ Scenario: Displaying an edit button on the latest ontology if it is failed with
   And I am logged in
   And there is an ontology with a "failed" version
   And I have permissions to edit the ontology
+  And there is a ontology file
```
eugenk: typo: there is *an* ontology file ;)

eugenk:
:+1:

PR #1651, opened 2016-04-16
1626 Remove duplicate method definition

### LINKED ISSUE #1626: Fix contradictory code in team_users
in [app/models/team_user.rb:17-21](https://github.com/ontohub/ontohub/blob/5d213a05436c75038479e0bcc0bcdf87e2005a05/app/models/team_user.rb#L17-L21) we have:

```ruby
  delegate :team_users, :to => :team

  def team_users
    self.class.where(:team_id => team_id)
  end
```

which makes two definitions for the same method name. Check which one is actually used and remove the other.

phyrog:
This fixes #1626

### DIFF
--- a/app/models/team_user.rb
+++ b/app/models/team_user.rb
@@ -16,10 +16,6 @@ class TeamUser < ActiveRecord::Base
 
   delegate :team_users, :to => :team
 
-  def team_users
-    self.class.where(:team_id => team_id)
-  end
-
   protected
 
   def set_admin_if_first_member


eugenk:
Can you change the `delegate` to a `has_many :team_users, through: :team` (maybe you need to specify the association on `Team` as well), please? This saves us one SQL query.

phyrog:
Done. It's not necessary to change `Team`, as far as I see.

eugenk:
It seems like there is a whole lot more than just this duplicate line wrong with `Team` and `TeamUser`:
There is a duplicate
```ruby
  attr_accessor   :admin_user
  attr_accessible :admin_user
```
in `team.rb` and the `team_user_spec.rb` uses lazy instantiation of the variables, but tests as if there was strict instantiation (`team_admin`, `non_admim_user` (yes, it has a typo) are only defined after their first call).

I will take a look at it a little more later today.

eugenk:
Let's keep the `delegate` - Rails can't handle a `has_many _, through _` where the `through`-part is also a `has_many` association.

eugenk:
:+1:

PR #1652, opened 2016-04-16
1631 Remove settings_for_development and fix hets paths

### LINKED ISSUE #1631: hets is not found when using hets-server
Ontohub depends in several places (e.g. https://github.com/ontohub/ontohub/blob/staging/spec/support/fixtures_generation/base_generator.rb#L10) on an executable called `hets`, which does not exist when using `hets-server`.

phyrog:
This fixes #1631. It essentially removes everything that has to do with `hets-lib` and `settings_for_development.yml` in favor of a `settings.local.yml`. It also fixes some hets paths.


lib/tasks/import.rake
```diff
@@ -7,38 +7,6 @@ namespace :import do
...
-
-      repo = Repository.new(name: 'Hets lib')
-
-      begin
-        repo.save!
-      rescue ActiveRecord::RecordInvalid
-        abort '"Hets lib" repository already existing.'
-      end
-
-      repo.import_ontologies(user, hets_lib_path)
```
eugenk: This is the only usage of the method `import_ontologies`. Please delete it (and `import_ontology`) in `app/models/repository/ontologies.rb` as well.

phyrog: Done.

eugenk:
:+1:

PR #1657, opened 2016-04-18
Set max error height to 75% of the display height

### LINKED ISSUE #1656: Limit error display height
When I visit the Versions tab of an ontology
And a version has failed
And the error is long
Then the height of the error-div should be limited to 75% of the window height.

But it is unlimited.

phyrog:
This should fix #1656


ebolloff:
:+1:

PR #1663, opened 2016-04-25
Content tab is not active

### LINKED ISSUE #1653: Content tab is not active
When I visit the page of a SingleOntology
And I click on "Content"
Then the content tab should be highlighted.

But it isn't.

ebolloff:
This should fix #1653.


eugenk:
:+1:

PR #1675, opened 2016-05-08
1610 offline hetsinstances get chosen and raise errors

### LINKED ISSUE #1610: Offline HetsInstances get chosen and raise errors
If the machine running a HetsInstance is offline and this instance gets chosen, an error is raised:
```
Errno::EHOSTUNREACH: No route to host - connect(2) for "madrid" port 800
```
The same instance gets chosen again and again. It must be marked as offline and another instance must be chosen.

phyrog:
This PR closes #1610  and does four things:

- set the `up` flag on `HetsInstance`s correctly when not reachable, so that only online `HetsInstance`s are chosen
- show the correct error message when no `HetsInstance` is available (`There is no HetsInstance which is reachable and has a minimal Hets version of v0.99, 1439380563`)
- schedule a retry job for parsing an ontology for an hour later if no `HetsInstance` is available
- remove already scheduled retries for a specific version when the user hits the `Retry` button, so the same version is only parsed once (unless again there is no `HetsInstance` available, in which case another retry is scheduled)

To achieve the third point I added a new gem `sidekiq-retries` which offers the ability to skip retries or do retries although the max amount of retries has already run. In this case I used it to run a retry only when a certain exception (`HetsInstance::NoSelectableHetsInstanceError`) was thrown.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -95,6 +95,7 @@ gem 'sidekiq', '~> 3.5.3'
 # The ref is given to ensure that no other (possibly breaking) changes are taken.
 gem 'sidetiq', github: 'PaulMest/sidetiq', ref: 'd88f9e483affcbadbd9e8b98b4a0a9518933887a'
 gem 'sidekiq-failures', '~> 0.4.5'
+gem 'sidekiq-retries', '~> 0.4.0'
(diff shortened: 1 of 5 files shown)


ebolloff:
:+1:

PR #1677, opened 2016-05-09
1612 handle concurrency of rugged and cli git

### LINKED ISSUE #1612: Handle concurrency of rugged and CLI git
Use a lock whenever something is done on a git repository to avoid race conditions between rugged and the command line tool git. Create/destroy the lock for git-ssh in a git hook.

phyrog:
This should fix #1612


phyrog:
I'll ignore the `Method has too many lines` messages since they refer to methods that were not written in this PR and should be addressed in a refactoring.

git/lib/git_update.rb
```diff
@@ -26,6 +26,12 @@ def initialize(repo_path, key_id, refs)
   end
 
   def exec
+    Semaphore.exclusively("action:#{@repo_name}") do
```
eugenk: Please find a way to only write `"action:"` once. Maybe put it into a constant or a class method of `GitRepository`.
And please use `git` instead of `action` (we will have many locks with different types and all of them are some kind of action)

eugenk:
I think you could incorporate it as well in lib/git_repository/cloning.rb:98 (`git_exec`) jsut to be sure.

eugenk:
See [spec/lib/semaphore_spec.rb](https://github.com/ontohub/ontohub/blob/b986f64d96c7ff863e249f1d8f7a47f4c4423c8d/spec/lib/semaphore_spec.rb#L11-L27) for a test that uses [fork_break](https://github.com/forkbreak/fork_break)

lib/git_repository/committing.rb
```diff
@@ -14,8 +14,22 @@ def add_file(userinfo, tmp_path, target_path, message, &block)
...
+      commit_file_without_lock(userinfo,
+                               file_contents,
+                               target_path,
+                               message,
+                               &block)
+    end
+  end
+
+  # change a single file and commit the change
+  def commit_file_without_lock(userinfo,
```
eugenk: This is not related to the PR, so don't mind.

git/lib/git_update.rb
```diff
@@ -26,6 +26,12 @@ def initialize(repo_path, key_id, refs)
...
+      exec_without_lock
+    end
+  end
+
+  def exec_without_lock
```
eugenk: This is not related to the PR, so don't mind.

eugenk:
Well, except for the yet to be written tests: :+1:

PR #1681, opened 2016-05-09
Change has_one to delegate

### LINKED ISSUE #1655: pagination for sentences does not work
With the ontologies here: http://ontohub.org/coinvent_meeting/house%20boat-new///children
when trying to display the second page (click on 2 or Next), I get a "The page you were looking for doesn't exist." error.

phyrog:
This should fix #1655. Note that it still does not work because the model `has_one :locid`, which depends on #1618. So this should probably wait until #1618 is merged.

- [x] Wait for #1618

### DIFF
--- a/app/models/translated_sentence.rb
+++ b/app/models/translated_sentence.rb
@@ -3,8 +3,9 @@ class TranslatedSentence < ActiveRecord::Base
   belongs_to :audience, class_name: Ontology.to_s
   belongs_to :sentence
   belongs_to :symbol_mapping
-  has_one :name, through: :sentence
   has_one :locid, through: :sentence
+
+  delegate :name, to: :sentence
 
   attr_accessible :audience, :ontology, :sentence, :symbol_mapping
   attr_accessible :translated_text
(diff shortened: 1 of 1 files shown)


phyrog:
So, turns out there are still problems with this pagination. We use [jQuery templates](https://github.com/BorisMoore/jquery-tmpl) to build and update the pagination view, but our content policy is too strict for that to work:

```
Uncaught EvalError: Refused to evaluate a string as JavaScript because 'unsafe-eval' is not an allowed source of script in the following Content Security Policy directive: "script-src 'self'".
```

So either we need to put the templates into separate files or we need to find another way to build the pagination.

Also note that jquery-templ has been superseeded by [jsrender](https://github.com/borismoore/jsrender), which seems to have the same problem.

Any thoughts on this, @ebolloff and @eugenk?

phyrog:
Problem from #1655 is fixed, as can be confirmed on https://develop.ontohub.org/sandbox/house%20boat-new//boat_house__T///axioms?all=true

The problem with jquery-templ still exists, but is hotfixed on develop by using a different content policy. This is not applicable on ontohub.org without risking XSS attacks.

ebolloff:
:+1:

PR #1683, opened 2016-05-09
Cannot edit user

### LINKED ISSUE #1185: cannot edit users
when I try to edit a user and then press "update user", I get an error.

ebolloff:
This should fix #1185 and also adds a test for it.


phyrog:
Updating the email address does not work. It just reverts to the previous one without an error.

ebolloff:
> Updating the email address does not work. It just reverts to the previous one without an error.

Changed with  efb1071

phyrog:
:+1: when tests are green

PR #1684, opened 2016-05-09
Set X-Frame-Options to SAMEORIGIN

### LINKED ISSUE #1632: sidekiq admin web interface is not loading
Commit 450c425d47cb4ff346c009bf6193057cee85fcc6 makes the sidekiq admin web interface disappear.

phyrog:
This fixes #1632. At the moment the ~~CSS~~ Javascript seems to be broken, though (the iframe is too small).

### DIFF
--- a/app/assets/javascripts/jobs-iframe.js.coffee
+++ b/app/assets/javascripts/jobs-iframe.js.coffee
@@ -1,0 +1,2 @@
+$("iframe#jobs").load ->
+  this.style.height = this.contentWindow.document.body.scrollHeight + 'px'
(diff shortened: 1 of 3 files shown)


ebolloff:
Please find a way to extend the sidekiq panel, even tough its not as beautiful as desired ;)

phyrog:
> Please find a way to extend the sidekiq panel, even tough its not as beautiful as desired ;)

Fixed. Was a similar problem than the IFrame issue. In this case the Content Security Policy forbids inline scripts.

ebolloff:
:+1:

PR #1689, opened 2016-05-11
Add sidekiq wrappers.

eugenk:
These are the most lightweight Sidekiq wrappers I can create.

```
Usage: bin/sidekiq_{default,hets,sequential} [options]
    -w, --workdir=MANDATORY          Set work directory
```

They create a pid file and a log file in the Workdir. The pid file is actually needed for the executable `bin/sidekiqctl`. It listens to the signals 
* `TERM` to terminate after at most 15 seconds (timeout parameter)
* `USR1` to terminate as soon as all running jobs have been processed
* `USR2` to reopen the log file

as described in the [corresponding Sidekiq wiki page](https://github.com/mperham/sidekiq/wiki/Signals)

On my laptop, `sidekiq_hets` takes 0.5 seconds to start up while the other two take 0.006 seconds to start up (i.e. to get to the `exec` command).

This is already deployed on tc.


eugenk:
Oh, and feel free to create a patch on the deployed instances to make a more specific shebang line, as you did with `git/bin/git-shell`

bin/sidekiq_wrapper_base.rb
```diff
@@ -0,0 +1,27 @@
...
+                            --environment #{ENV['RAILS_ENV'] || 'production'}
+                            --timeout 15)
+
+# Parse options
+def parse_options(filename)
+  options = {}
+  OptionParser.new do |opts|
+    opts.banner = "Usage: #{filename} [options]"
+
+    opts.on('-wMANDATORY', '--workdir=MANDATORY', 'Set work directory') do |w|
```
jelmd: making _options_ mandatory is almost in any case bad design. Use CDW if not given.

eugenk: Done with f5bf00c

jelmd:
Don't understand, why there need to be 3 files for the same? purpose.
Which one should be taken?
The timeout should be a param as well, so that it can be changed, when more knowledge has been collected...

eugenk:
> Don't understand, why there need to be 3 files for the same? purpose.
Which one should be taken?

We need three different sidekiq instances each with different parameters. For each of these instances, I created a wrapper so that the init script can be very short. The wrappers are subject to change because the parameters may need some tweaking in the future (queue priorities, concurrency).

> The timeout should be a param as well, so that it can be changed, when more knowledge has been collected...

Done with 5798b5c

eugenk:
> Which one should be taken?

All three sidekiq instances need to be started in parallel.

jelmd:
No - this is a no go. This way the application logic is pushed back to the starter script which is wrong, buys us nothing (especially because Linux has no contract filesystem for processes).

As said, the init script starts __one__ process and this one does, whatever is needed and daemonizes itself. It should provide required flexibility by accepting related options (if not given, choose reasonable defaults). And this one is the one where to send the signals. It than can intercept, xlate or broadcast to forked children or do whatever the documentation says ;-)

BTW: Since all the files do more or less the same, have the same purpose, they belong IMHO into a single file. See it as a Starter class, which has certain properties/different starter methods ...

eugenk:
Alright, now there's only one wrapper to start and stop all instances of sidekiq. It delegates the above-mentioned signals to them.

jelmd:
Okidoki, I'll try it ... :)

bin/sidekiq_wrapper
```diff
@@ -5,17 +5,20 @@ require_relative '../lib/environment_light.rb'
...
   OptionParser.new do |opts|
     opts.banner = "Usage: #{File.basename(__FILE__)} [options]"
 
+    opts.on('-eMANDATORY', '--environment=MANDATORY',
+            "Set Rails environment (default: #{options[:environment]})") do |e|
+      options[:environment] = e
+    end
+
     opts.on('-wMANDATORY', '--workdir=MANDATORY',
             'Set work directory (default: CWD)') do |w|
```
jelmd: BTW: dict lookup __MANDATORY__!

Also as a general rule of thumb: In synopsis, usage infos the option value names should indicate its "type", e.g. -w *directory*, -t *seconds*. The description should just include '(mandatory)' or '(optional)' if needed. Usually it is not, because the default value should be mentioned and than it is clear, that it is 'optional'. For arg-less options the name _option_ indicates, that it is _option_al and thus doesn't need to be explicitly mentioned. So only in very rare cases or combination of options may require, that 'mandatory' appears in documentation.

eugenk: The `MANDATORY` is Ruby's `optparse`'s way to say that **if** `-w` is used, **then** a value must be specified. This does not say that you have to use `-w` at all.

jelmd:
The service (i.e. sidekiq_wrapper) does not daemonize itself, but every service always should after it has validated config[s] and runability (started needed workers) successfully. If not, it should exit wqith a proper exit code, which let the "restarter" determine, whether it makes sense to try to restart, whether there is a general error (e.g. missing disk ;-)) or config error (and thus needs to inform the user), etc..

Something like http://daemons.rubyforge.org/Daemonize.html may help ... ?

eugenk:
Okay. I will implement it, but not today. Preumably on tuesday.

bin/sidekiq_wrapper
jelmd: Admins usually have no notion of ill/senseless optparse documentation. However, they know [indirectly] http://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap12.html by reading man pages and usage synopsis from native Unix utilities ...

eugenk:
You can now daemonise the process with a separate script (here with all parameters exemplary specified):

```
bin/sidekiq_wrapper_daemon start -- -e production -w /tmp/sidekiq -t 15
bin/sidekiq_wrapper_daemon stop
bin/sidekiq_wrapper_daemon restart
```

This script calls `bin/sidekiq_wrapper` and then daemonises itself. I couldn't get it to work with only one script.

eugenk:
Oh, and the daemon process puts a pidfile of itself to `$RAILS_ROOT/tmp/pids`. Unfortunately, the `daemons` gem is not quite compatible to the options parser, so I can't get the `-w` to work for the daemon easily. I will try again, but I can't guarantee it.

eugenk:
We need to wait for https://github.com/spechub/Hets/issues/1644 to get this running.

eugenk:
@jelmd: https://github.com/spechub/Hets/issues/1644 has been merged, so with the next version of Hets, you can specify a pidfile with `hets -X -P path_to_pidfile`.
Please do so in the svcadm scripts and post in this issue where (all) the pidfiles are located.

Issue #1691, opened 2016-05-17
Upgrade capybara gem


eugenk:
Upgrading capybara results in failing features. Resolve the problems.

Issue #1690, opened 2016-05-17
Upgrade secure_headers gem


eugenk:
Upgrade secure_headers to the current version and configure its new options reasonably. See the changelog at https://github.com/twitter/secureheaders/releases

Issue #1686, opened 2016-05-09
Replace inherited_resources by responders gem


phyrog:
[josevalim/inherited_resources](https://github.com/plataformatec/responders) is no longer maintained and should be replaced by [plataformatec/responders](https://github.com/plataformatec/responders). responders requires Rails version 4.2+

Issue #1340, opened 2015-05-05
Update inherited_resources


eugenk:
[Inherited resources](https://github.com/josevalim/inherited_resources) version 1.5.0 contains https://github.com/josevalim/inherited_resources/commit/cfb2d88c6d93dbaf9bd4a81996a660181fd2b2e1, which breaks our combinations feature.

To reproduce the error:
* update inherited resources to the current version (currently 1.6.0)
* run `rake cucumber`
  * some action- and some combination-features are going to be red, because `association_chain` is being called which in turn calls `find_by_path!` on `nil`.

I suspect the following lines of our `app/controllers/api/v1/combinations_controller.rb` to be the bad guys.
```ruby
  belongs_to :repository, finder: :find_by_path!
  belongs_to :ontology, optional: true
```

PR #1692, opened 2016-05-17
Update gems

eugenk:
This updates most of our used gems that are compatible to rails 3.2

The only gems that are not included here, are:
* inherited_resources (see #1340 and #1686)
* secure_headers (see #1690)
* capybara (see #1691)
* rspec (see #1693)

Every time `rake` is invoked, we get
```
[DEPRECATION] `last_comment` is deprecated.  Please use `last_description` instead.
```
rspec 3.4.4 fixes this.


eugenk:
Jenkins, test this please.

ebolloff:
:+1:

PR #1695, opened 2016-05-20
Remove .ruby-version-deployment.

eugenk:
This removes the file .ruby-version-deployment because we don't want to use it any more.


jelmd:
+1

PR #1730, opened 2016-07-08
Hotfix deadlocks in case of "no imports from one file into many".

eugenk:
This shall hot fix deadlocks when there are "no imports from one file into many". It is a dirty fix for #1642. An issue for a proper fix is filed with #1729.

PR #1732, opened 2016-07-09
Fix deadlocks caused by semaphore.

### LINKED ISSUE #1729: Fix Deadlocks Caused by Semaphore
This is the successor of #1642. Jobs are not performed because of deadlocks that are caused by our Semaphore class. See https://github.com/ontohub/ontohub/issues/1642#issuecomment-231241099 for a description of the problem. The same problem occurs, for example, in [lib/hets/concurrent_evaluator.rb:39](https://github.com/ontohub/ontohub/blob/329a411da882f3d451b3257eea99c9262b4c3a68/lib/hets/concurrent_evaluator.rb#L39) if it is called multiple times in parallel.

A proper fix needs to change the gem [redis-semaphore](https://github.com/dv/redis-semaphore) in a way that `locked?` can be used across different processes (hence, on different semaphores on the same key). Instead of our complicated Semaphore class (which is complicated because of our implementation of `locked?`), we should only wrap the original redis-semaphore.

eugenk:
This shall fix #1729 and supersede #1730.

With my fork of the redis-semaphore gem, we can simply use the methods of the gem instead of implementing locking on top of it. There, the `locked?` method is fixed.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -15,7 +15,9 @@ gem 'rdf', '~> 1.99.0'
 gem 'rdf-rdfxml', '~> 1.99.0'
 gem 'rdf-n3', '~> 1.99.0'
 
-gem 'redis-semaphore', '~> 0.3.1'
+# As soon as the pull request https://github.com/dv/redis-semaphore/pull/46 is
+# merged and a new release is out, use the upstream gem again.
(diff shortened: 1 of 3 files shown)


ebolloff:
:+1:

PR #1734, opened 2016-07-19
Add puma_worker_killer gem

phyrog:
Adds the [puma worker killer](https://github.com/schneems/puma_worker_killer) gem to keep memory usage low.

Should be tested on develop before merging.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -195,6 +195,7 @@ group :production do
   # puma is __the only exception__ for which we don't specify a version.
   gem 'puma'
   gem 'eye', '~> 0.8'
+  gem 'puma_worker_killer', '~> 0.0.6'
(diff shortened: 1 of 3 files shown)


phyrog:
This is deployed on develop now. The best way to check if it's working is probably to use develop and check after some hours/days if the memory usage has increased significantly. By default the workers are restarted on a rolling basis every 6 hours.

ebolloff:
I clicked some large ontologies on `develop`. Currently the memory is about 8GB. I will check later, if the killer worked properly.

phyrog:
The failure is unrelated to this PR, but I created a new issue for it (#1736).

ebolloff:
:+1:

PR #1737, opened 2016-07-19
Index specific_* fields in loc_ids table

eugenk:
This significantly speeds up page loading. For example, the landing page took some 1000 ms to load before and now it takes 275 ms. And the file browser view in bioportal took about 94 seconds (!) which slimmed down to 34 seconds.


ebolloff:
:+1:

PR #1739, opened 2016-07-19
Preload locids for all models

### LINKED ISSUE #1738: Improve loc_id performance
In a lot of cases, the locid of something is queried in a separate query. By preloading the locid for all models that have a locid we can hopefully improve the overall performance (e.g. every use of `LinkHelper#fancy_link` requires the locid, so for a list of 3000 things with links, we currently need 1+3000 queries).

phyrog:
This should fix #1738. This partly depends on #1681.

For this to work, we need to migrate some data (Sentences and subclasses), so a very careful review is needed.

- [x] write data migration

### DIFF
--- a/app/models/loc_id_base_model.rb
+++ b/app/models/loc_id_base_model.rb
@@ -6,7 +6,7 @@ class LocIdBaseModel < ActiveRecord::Base
   # There is no polymorphic has_one
   has_many :loc_ids, as: :specific
 
-  default_scope { includes(:loc_ids) }
+  default_scope { joins(:loc_ids) }
 
   def self.find_with_locid(locid, iri = nil)
     result = LocId.where(locid: locid).first.try(:specific)
--- a/app/models/loc_id_base_model.rb
+++ b/app/models/loc_id_base_model.rb
@@ -52,7 +52,7 @@ class LocIdBaseModel < ActiveRecord::Base
     # Sometimes the objects are "Ontology" and sometimes a subclass.
     if [DistributedOntology, SingleOntology].include?(self.class)
       Ontology
-    elsif [Axiom, Theorem].include?(self.class)
+    elsif Sentence.descendants.include?(self.class)
       Sentence
     else
       self.class
--- a/app/models/loc_id_base_model.rb
+++ b/app/models/loc_id_base_model.rb
@@ -3,6 +3,11 @@ class LocIdBaseModel < ActiveRecord::Base
   after_create :create_locid
   before_destroy :destroy_locid
 
+  # There is no polymorphic has_one
+  has_many :loc_ids, as: :specific
+
+  default_scope { includes(:loc_ids) }
+
   def self.find_with_locid(locid, iri = nil)
     result = LocId.where(locid: locid).first.try(:specific)
     if table_name == 'ontologies' && result.nil? && iri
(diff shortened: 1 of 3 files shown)


app/models/loc_id_base_model.rb
```diff
@@ -43,15 +48,12 @@ def locid=(string)
...
-  def query_locid
-    LocId.where(specific_id: id,
-                specific_type: normalized_class.to_s)
-  end
-
   def normalized_class
     # Sometimes the objects are "Ontology" and sometimes a subclass.
     if [DistributedOntology, SingleOntology].include?(self.class)
       Ontology
+    elsif [Axiom, Theorem].include?(self.class)
```
eugenk: Maybe we could use `if Sentence.descendants.include?(self.class)` at this point.

The same does not work for `Ontology`, at least on my machine: 
```ruby
[1] pry(main)> Ontology.descendants
=> []
[2] pry(main)> Sentence.descendants
=> [Axiom(id: integer, ontology_id: integer, name: text, text: text, range: string, comments_count: integer, created_at: datetime, updated_at: datetime, display_text: text, imported: boolean, type: string, proof_status_id: string, state: string, state_updated_at: datetime, last_error: text, provable: boolean),
 Theorem(id: integer, ontology_id: integer, name: text, text: text, range: string, comments_count: integer, created_at: datetime, updated_at: datetime, display_text: text, imported: boolean, type: string, proof_status_id: string, state: string, state_updated_at: datetime, last_error: text, provable: boolean)]
```
I wonder why this is.

db/data/20160719133546_preload_locids.rb
```diff
@@ -0,0 +1,19 @@
+class PreloadLocids < ActiveRecord::Migration
+  CLASSES = [Axiom, Theorem].freeze
+
+  def self.up
+    LocId.where(specific_type: CLASSES.map(&:to_s)).find_each do |object|
```
eugenk: Here, you could use `Sentence.descendants` instead of `CLASSES`, as well. (more descriptive)

db/data/20160719133546_preload_locids.rb
```diff
@@ -0,0 +1,17 @@
+class PreloadLocids < ActiveRecord::Migration
+  def self.up
+    LocId.where(specific_type: Sentence.descendants.map(&:to_s)).find_each do |object|
```
phyrog: Since this is just a migration I think these 6 chars are OK.

phyrog:
After testing this on develop, I now believe preloading locids for all models is a bad idea.

Some metrics for `/mappings`, which is a rather large list with 3000 entries:

Without #1739: ~15000ms
With #1739: ~25000ms
With #1739 and further optimizations: ~5000-8000ms
Without #1739 and with further optimizations: ~4000ms

The optimizations were `Mapping.joins(:ontology).joins(:symbol_mappings)`

Interestingly, `joins(:loc_ids)` uses exactly as many queries as not using it (~2300 in mappings). The other optimizations bring the query count down to ~700.

So instead of preloading locids everywhere, we should decide from case to case where to preload what. Apart from that, we should check which of our utility functions generate useless queries (I found one that queries `COUNT(*)` right before actually querying everything).

What I would like to keep from this PR is everything apart from the `default_scope`.

eugenk:
> Interestingly, `joins(:loc_ids)` uses exactly as many queries as not using it (~2300 in mappings). The other optimizations bring the query count down to ~700.

That's because `joins` only creates a `JOIN` SQL-statement, but does not populate the ActiveRecord objects that can be accessed by the corresponding association. As soon as you call, for example
```ruby
ontology = Ontolgy.joins(:loc_id).find(1)
ontology.locid
```
Rails sends another `SELECT * from loc_ids WHERE specific_type = Ontology AND specific_id = 1`.

If you use `includes` instead of `joins`, then Rails already populates the `ontology.loc_id` object and with it, also the string `ontology.locid`.

eugenk:
> So instead of preloading locids everywhere, we should decide from case to case where to preload what.

Good point.

phyrog:
Oh you're right. I assumed that join also populates the associations. I will try using `eager_load`, which seems to do a join with populating the associations.

phyrog:
Closed in favour of #1766

Issue #1736, opened 2016-07-19
Commit saving a commit with author, committer and pusher being the same user saves the committer_name


phyrog:
I got this failure on an unrelated PR.

```
1) Commit saving a commit with author, committer and pusher being the same user saves the committer_name
     Failure/Error: expect(subject.send("#{field}_name")).to eq(pusher.name)
       
       expected: "Ole Kertzmann Jr."
            got: "Ole Kertzmann Jr"
       
       (compared using ==)
     # ./spec/models/commit_spec.rb:37:in `block (5 levels) in <top (required)>'
```

phyrog:
It seems, that rugged somehow strips the "." at the end. As far as I can see, there is nothing we can do about it in Ontohub (apart from changing the test).

PR #1740, opened 2016-07-19
Remove profiler gem

eugenk:
This disables profiling because it violates the content policy.

The profiling is active on develop (**only** on develop) with 617ccf03306fd8b9770859474d7a869072d1418f


phyrog:
For the test fail see #1736. It's unrelated to this PR, so :+1:

Issue #1555, opened 2016-01-06
Links in details of ontology-mapping graphs are wrong


eugenk:
For example, in http://localhost:3000/default/partial_order//v__T///graphs after clicking on a node. Translate the links to locids.

Maybe wait for #1450 to finish. The `link_to` method could be changed to use locids for objects of the locid-defining class.

PR #1744, opened 2016-07-26
1555 links in details of ontology mapping graphs are wrong

eugenk:
This shall fix the URL-bug of #1555. It also fixes another URL-bug for the nodes of LogicMappings.


ebolloff:
:+1:

PR #1745, opened 2016-07-26
1555 patch url for

eugenk:
This shall fix the `url_for` part of #1555.

Now, Rails's `url_for` replaces our `locid_for` to make our code more streamlined.

The monkey patches we use are now documented as well.

Issue #1458, opened 2015-08-18
loc/ids are not used everywhere


tillmo:
in particular, after uploading an ontology via the file browser, or after starting a proof of a theorem, I can see numbers in my browser's address field.

eugenk:
This is a huge task. We would need to create additional routes for all actions on controllers using LocIdBaseModel-extending classes. I think the best way would be to change the router itself, so that `resources` recognizes if the resource is of a LocIdBaseModel-extending class and then handles this specifically.

I don't know, though, if this is feasible. The other option is, as said, to define `specified_get`, `specified_post` etc routes for all actions that are missing it.

eugenk:
Remaining task: Implement routes for all imaginable Loc/Ids.

PR #1746, opened 2016-07-26
1458 loc ids are not used everywhere

eugenk:
This shall fix only the two cases mentioned in #1458. It gives a glimpse of what needs to be changed to cover everything.

A proper fix could be to implement a better suitable router instead of using Specroutes (which is incompatible to Rails 4.2 anyway).

--

~~This branch is based on `1555-patch_url_for` of #1745 and **not based on staging**. It may be easier to review that one first.~~

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -185,6 +185,32 @@ Will provide a subsite of a specific ontology.
       end
   end
 
+  specified_get '/:repository_id/*locid///edit' => 'ontologies#edit',
+    as: :ontology_edit,
+    constraints: [
+      LocIdRouterConstraint.new(Ontology, ontology: :id),
+    ] do
+      accept 'text/html'
+
+      doc title: 'Ontology edit command',
+          body: <<-BODY
+Will provide a site to the ontology command.
+      BODY
+    end
+
+  specified_put '/:repository_id/*locid' => 'ontologies#update',
+    as: :ontology_update,
+    constraints: [
+      LocIdRouterConstraint.new(Ontology, ontology: :id),
+    ] do
+      accept 'text/html'
+
+      doc title: 'Ontology update command',
+          body: <<-BODY
+Will provide a site to the ontology command.
+      BODY
+    end
+
   specified_get "/:repository_id/*locid///sentences" => "api/v1/sentences#index",
     as: :"ontology_iri_sentences",
     constraints: [
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -141,6 +141,19 @@ MMT-query-string.
       BODY
     end
 
+  specified_post '/:repository_id/*locid///retry' => 'ontologies#retry_failed',
+    as: :ontology_retry,
+    constraints: [
+      LocIdRouterConstraint.new(Ontology, ontology: :id),
+    ] do
+      accept 'text/html'
+
+      doc title: 'Ontology retry parsing command',
+          body: <<-BODY
+Will parse the ontology again.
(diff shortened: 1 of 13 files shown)


lib/extend/action_dispatch/routing/polymorphic_routes.rb
```diff
@@ -0,0 +1,49 @@
...
+              record_or_hash_or_array.pop
+            else
+              {}
+            end
+
+          commands = record_or_hash_or_array
+
+          build_locid_url(base_url, record, commands, query_components)
+        else
+          original_definition_of_polymorphic_url.
```
eugenk: This is part of #1745 and already fixed there.

config/routes.rb
```diff
@@ -265,8 +345,36 @@
       end
   end
 
+  specified_post '/:repository_id/*locid///prove' => 'proofs#create',
+    as: :"theorem_prove",
```
eugenk: This is fixed by rebase + commit-amend.

config/routes.rb
```diff
@@ -169,6 +169,34 @@
...
+    constraints: [
```
eugenk: I would like to leave it as is because it's more readable.

config/routes.rb
eugenk: Alignment: I would like to leave it as is because it's more readable.
Strings: This is fixed by rebase + commit-amend.

eugenk: Alignment: I would like to leave it as is because it's more readable.
Strings: This is fixed by rebase + commit-amend.

config/routes.rb
```diff
@@ -265,8 +345,36 @@
       end
   end
 
+  specified_post '/:repository_id/*locid///prove' => 'proofs#create',
+    as: :theorem_prove,
```
eugenk: I would like to leave it as is because it's more readable.

config/routes.rb
```diff
@@ -169,6 +169,34 @@
...
+      doc title: 'loc/id reference to a theorem command',
+          body: <<-BODY
+Will return a representation of the theorem command. The theorem
+is determined according to the *locid.
+      BODY
+    end
+
+  specified_get '/:repository_id/*locid///prove' => 'proofs#new',
+    as: :theorem_prove,
+    constraints: [
```
eugenk: I would like to leave it as is because it's more readable.

config/routes.rb
```diff
@@ -169,6 +169,34 @@
...
+
+      doc title: 'loc/id reference to a theorem command',
+          body: <<-BODY
+Will return a representation of the theorem command. The theorem
+is determined according to the *locid.
+      BODY
+    end
+
+  specified_get '/:repository_id/*locid///prove' => 'proofs#new',
+    as: :theorem_prove,
```
eugenk: I would like to leave it as is because it's more readable.

ebolloff:
We have the `Sentence_and_symbol_with_same_name` ontology in our seeds. When I click on it and on the "Retry" button after that I get this iri `http://localhost:3000/repositories/default/ontologies/7/versions`

eugenk:
This is done with a3c387c549f366e54cd9bab66c5685a3178a09ff

eugenk:
I think we can ignore Hound because of the special structure of the specroutes

ebolloff:
We should add the `routes.rb` to the excluded files of Hound. I guess @phyrog will do it in #1768. :+1:

PR #1747, opened 2016-07-27
1719 test suite does not honor hets port settings

### LINKED ISSUE #1719: test suite does not honor hets port settings
When 'bundle exec rake' gets run, it doesn't honor hets port settings from config/settings.local.yml but uses port 8000 always. This may break tests, if another instance is running at the same time.

E.g. see http://ohj.iks.cs.ovgu.de:8080/job/ontohub-PRs/37/consoleFull (search for 8000).

eugenk:
This shall fix #1719 by adding a new settings key for the port that Hets shall listen to when generating the test fixtures. We cannot use the `instance_urls` key for this because these instances may all run on other hosts.

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -92,6 +92,10 @@ hets:
   # This is the path to the hets executable we use in `rake hets:*` and for the
   # process manager in production mode (god)
   executable_path: /usr/bin/hets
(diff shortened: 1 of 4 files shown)


ebolloff:
:+1:

PR #1748, opened 2016-07-27
1620 deletion of child ontologies

### LINKED ISSUE #1620: Deletion of child ontologies
We have no defined behaviour of the deletion of child ontologies.

Scenario:
When I upload a distributed ontology D with children C1, C2, ..., Cn
And I upload a new version of D1 with children C1, C2, ... C(i-1), C(i+1), ..., Cn
Then Ci should be inactive in the newer version, but active in an older version.

Currently, this makes our application fail (see #1593).

eugenk:
This shall fix #1620.

--

~~This branch is based on `1555-patch_url_for` of #1745 and **not based on staging**. It may be easier to review that one first.~~

### DIFF
--- a/app/models/ontology/associations_and_attributes.rb
+++ b/app/models/ontology/associations_and_attributes.rb
@@ -2,6 +2,11 @@ module Ontology::AssociationsAndAttributes
   extend ActiveSupport::Concern
 
   included do
+    acts_as_tree
+    # Add the scope "present: true" to the children association by acts_as_tree.
+    Ontology.reflect_on_association(:children).
+      options[:conditions] = {present: true}
+
     belongs_to :language
     belongs_to :logic, counter_cache: true
     belongs_to :ontology_type
--- a/app/models/ontology/associations_and_attributes.rb
+++ b/app/models/ontology/associations_and_attributes.rb
@@ -6,6 +6,7 @@ module Ontology::AssociationsAndAttributes
     # Add the scope "present: true" to the children association by acts_as_tree.
     Ontology.reflect_on_association(:children).
       options[:conditions] = {present: true}
+    has_many :all_children, class_name: Ontology.to_s, foreign_key: 'parent_id'
 
     belongs_to :language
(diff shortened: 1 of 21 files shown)


app/models/ontology/associations_and_attributes.rb
```diff
@@ -2,6 +2,11 @@ module Ontology::AssociationsAndAttributes
   extend ActiveSupport::Concern
 
   included do
+    acts_as_tree
+    # Add the scope "present: true" to the children association by acts_as_tree.
+    Ontology.reflect_on_association(:children).
```
eugenk: This is done (rebase, commit amend).

eugenk: This is done (rebase, commit amend).

lib/hets/dg/node_evaluation_helper.rb
```diff
@@ -149,6 +149,16 @@ def update_ontologies_per_logic_count!(ontologies)
           Logic.reset_counters(logic_id, :ontologies)
         end
       end
+
+      def set_all_children_inactive_except(present_ontologies)
```
eugenk: No, thanks.

ebolloff:
When I update an ontology and update it to the initial version after that I get the following error

```
PG::UniqueViolation: FEHLER:  doppelter Schlüsselwert verletzt Unique-Constraint „index_loc_ids_on_locid“
DETAIL:  Schlüssel „(locid)=(/default/test2//sp1)“ existiert bereits.
: INSERT INTO "loc_ids" ("created_at", "locid", "specific_id", "specific_type", "updated_at") VALUES ($1, $2, $3, $4, $5) RETURNING "id"
```

ebolloff:
The error still occurs.

eugenk:
@ebolloff, this should work now.

ebolloff:
The tests hang up on my machine, but they work on @phyrog and @eugenk machines so :+1:

PR #1750, opened 2016-08-02
1707 pagination in ontology mappings is broken

### LINKED ISSUE #1707: Pagination in ontology mapping is broken
e.g. https://ontohub.org/sandbox/animal///mappings

- When clicking any of the pagination buttons, the wrong url is opened (`/sandbox/animal/mappings`).
- The ontology_id is appended as a query parameter, which seems to be unnecessary

Confirmable on master, develop and staging.

phyrog:
This fixes #1707. It also fixes the related global mappings list (`/mappings`), which also had pagination issues.

### DIFF
--- a/app/models/mapping.rb
+++ b/app/models/mapping.rb
@@ -37,11 +37,10 @@ class Mapping < LocIdBaseModel
                   :target_id, :versions_attributes, :versions, :name
   accepts_nested_attributes_for :versions
 
-
-  def self.with_ontology_reference(ontology_id)
-    Mapping.where('ontology_id = ? OR source_id = ? OR target_id = ?',
-                          ontology_id, ontology_id, ontology_id)
-  end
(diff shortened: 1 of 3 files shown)


app/models/mapping.rb
```diff
@@ -37,11 +37,10 @@ def current
                   :target_id, :versions_attributes, :versions, :name
   accepts_nested_attributes_for :versions
 
-
-  def self.with_ontology_reference(ontology_id)
-    Mapping.where('ontology_id = ? OR source_id = ? OR target_id = ?',
-                          ontology_id, ontology_id, ontology_id)
-  end
+  scope :with_ontology_reference, -> (ontology_id) {
+    where('ontology_id = ? OR source_id = ? OR target_id = ?',
```
phyrog: Fixed in a283f85

ebolloff:
Pagination raises errors often, so we need to fix it globally in the future. :+1:

PR #1755, opened 2016-08-02
Check if returned value is an Ontology

### LINKED ISSUE #1749: undefined method `symbols'
see https://develop.ontohub.org/sandbox/artih///ontology_versions

phyrog:
This should fix #1749. The returned object was not necessarily an ontology.


phyrog:
This fix is deployed on develop now.

ebolloff:
:+1: 
I don't get the symbols error, but this one
```
The name "1" of the symbol is ambiguous.
It is already used for either a axiom or a theorem.
The symbol "1" is located at /opt/hets-lib/Basic/Numbers.casl:133.12.
```
I guess, this needs a fix in hets?

phyrog:
> I guess, this needs a fix in hets?

I believe so. Is that correct, @tillmo, @eugenk?

eugenk:
This is definitely because the ontology is incompatible to Ontohub because two different objects would have the same Loc/Id.

PR #1756, opened 2016-08-02
"Fix" the pusher/committer test

### LINKED ISSUE #1736: Commit saving a commit with author, committer and pusher being the same user saves the committer_name

phyrog:
Should (hot)fix #1736. This is clearly not the optimal solution, but as far as I can see the most feasible one.

Can be checked by using an explicit user name like this:

```
let(:pusher) { create :user, name: "Mr. Johnson Jr." }
```

### DIFF
--- a/spec/models/commit_spec.rb
+++ b/spec/models/commit_spec.rb
@@ -34,7 +34,7 @@ describe Commit do
         end
 
         it "saves the #{field}_name" do
-          expect(subject.send("#{field}_name")).to eq(pusher.name)
+          expect(subject.send("#{field}_name").sub(/\.$/, '')).to eq(pusher.name.sub(/\.$/, ''))
         end
       end
 


eugenk:
:+1:

eugenk:
In this case, it doesn't matter, but `$` matches only the end of the line. You match for the end of the string with `\z`. As we only deal with a single line, `$` is just fine, at this point.

PR #1757, opened 2016-08-03
1753 wait for sql to finish before parsing worker

### LINKED ISSUE #1753: wait for sql to finish before parsing worker
When a new ontology is parsed, I get
```
ActiveRecord::RecordNotFound: Couldn't find OntologyVersion with id=...
```
The problem is that the Sidekiq parsing worker is started before the ontology version has been entered into the database.

Unfortunately, #1733 does not fix this problem.
A proper fix needs to ask Postgres whether the (probably asynchronously started) database transaction has already been finished. If needed, use sql code for this

eugenk:
This shall fix #1753 by adding a gem that allows to wait for all transactions to finish.

It is not feasible to use a `after_commit :method, on: :create` hook because this would save the version dozens of times and even create many versions. I could not find out why this happens. The gem, however, solves the problem.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -8,6 +8,7 @@ gem 'pry-rails', '~> 0.3.2'
 
 gem 'pg', '~> 0.18.1'
 gem 'foreigner', '~> 1.7.2'
+gem 'ar_after_transaction', '~> 0.4.0'
 
 gem 'rdf', '~> 1.99.0'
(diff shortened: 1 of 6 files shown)


eugenk:
Seems like this causes more trouble with the tests than it benefits... maybe there is a better solution.

eugenk:
Closing this unmerged in favour of #1761.

Issue #1578, opened 2016-01-26
Make OWL API work with Ontohub URLs that have no filename extension


tillmo:
For Ontohub URLs without filename extension, e.g. `http://ontohub.org/sandbox/inconsistent`, we correctly deliver the raw text if the accept header is set correctly, e.g.
```
wget --header "Accept: text/plain" http://ontohub.org/sandbox/inconsistent
```
However, the OWL API appearantly does not set the accept header in this way. As a consequence, Ontohub delivers html, and the OWL API throws a syntax error. You can observe this behaviour e.g. if you enter `http://ontohub.org/sandbox/inconsistent` into Protégé (menu: open ontology from URL).

eugenk:
The browser should send `text/html`. If no accept header is set, we should default to `text/plain`

tillmo:
will possibly be fixed by #1658, so wait for that first.

phyrog:
I believe this is a nofix like #1699? The new version of Protege (and the OWL API) does not seem to have this problem.

tillmo:
yes, we should refer users to the latest version of Protégé and OWL API.

phyrog:
I think my earlier comment was too hasty and we should re-open this issue. I just tried importing an ontology in Protege, and it does not throw an error, but it just fails silently and nothing is imported (which makes sense, how should Protege or the OWL API know how to parse our HTML).

What the new OWL API and Protege did fix, was the redirects, which did not work before; this case is something different entirely.

phyrog:
@0robustus1 Would this work when the `Accept` header is not present?

```
reroute_on_header(header: 'Accept', value: '', to: ...)
```

0robustus1:
Well it shouldn't fail if you are talking about that. But empty string won't match for non-existence of header. So it won't have the same semantic as `present?`. (see https://github.com/0robustus1/specroutes/blob/60e0b10eda4ede044e06af403eb494a994d1df95/lib/specroutes/constraints/header_constraint.rb#L19-L21 for more info).

0robustus1:
But we can change the behaviour in specroutes if you think something else is more feasible. Just ping me again with the wanted change, and we can do it ;)

phyrog:
Found another (easier) way around this by switching the `accept`s and the `reroutes`, which seems to work fine. Thanks nonetheless.

PR #1761, opened 2016-08-08
1753 wait for sql to finish before parsing worker retry

### LINKED ISSUE #1753: wait for sql to finish before parsing worker

eugenk:
This shall fix #1753 and replace #1757.

It removes the `after_create :async_parse` callback from OntologyVersion, because we don't need it. Instead, it schedules a parsing job in the sequential code, when needed.

As a consequence, we do not need the `do_not_parse` flag this much: We have used it to only parse the file once in case of a distributed ontology (and to not parse it for each child ontology). Now, we only use it in the seeds to be explicit and in pushes to call a different ParsingWorker.

### DIFF
--- a/lib/ontology_saver.rb
+++ b/lib/ontology_saver.rb
@@ -7,7 +7,7 @@ class OntologySaver
     self.repository = repository
   end
 
-  def save_ontology(commit_oid, ontology_version_options, changed_files: [])
+  def save_ontology(commit_oid, ontology_version_options, changed_files: [], do_parse: true)
     # We expect that this method is only called, when we can expect an ontology
     # in this file.
     file_extension = File.extname(ontology_version_options.filepath)
@@ -23,7 +23,7 @@ class OntologySaver
     ontology.present = true
     ontology.save!
 
-    async_parse_version(version) unless ontology_version_options.do_not_parse
+    async_parse_version(version) if do_parse
 
     version
   end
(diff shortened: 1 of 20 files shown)


lib/ontology_saver.rb
```diff
@@ -76,6 +78,22 @@ def mark_ontology_as_having_file(path, has_file: false)
     end
   end
 
+  def async_parse_version(version)
```
eugenk: I'm not sure if we should change this method. I let the reviewer decide.

lib/ontology_saver.rb
```diff
@@ -7,7 +7,8 @@ def initialize(repository)
     self.repository = repository
   end
 
-  def save_ontology(commit_oid, ontology_version_options, changed_files: [])
+  def save_ontology(commit_oid, ontology_version_options, changed_files: [],
```
eugenk: I don't know how to shorten the method reasonably.

lib/ontology_saver.rb
phyrog: Since the method is pretty straight forward, I think it's ok.

ebolloff: :+1:

ebolloff:
It works fine, but some ontologies (e.g. `generations.owl` in our seeds) fail because of #1578.

phyrog:
:+1: from me, parsing problems are unrelated to this PR.

PR #1766, opened 2016-08-08
1765 loc id base model associations are not complete

### LINKED ISSUE #1765: loc_id_base_model associations are not complete
This is basically a new issue for #1739 with a different focus (without the default scope). 

`loc_id_base_model` should `has_one :loc_id`. Unfortunately there is no polymorphic `has_one`, so we should use `has_many ..., as: ...` instead. This would simplify multiple methods in `loc_id_base_model`.

phyrog:
Fixes #1765. Same as #1739 apart from the default scope.


eugenk:
:+1:

PR #1768, opened 2016-08-10
1578 make owl api work with ontohub urls that have no filename extension

### LINKED ISSUE #1578: Make OWL API work with Ontohub URLs that have no filename extension

phyrog:
Should fix #1578.

The route now defaults to the API controller, unless a `Accept: text/html` is sent (which all browsers do).

Also `text/plain` is now the default format in the ontology API, unless a `Accept: application/json` is sent.

### DIFF
--- a/app/controllers/api/v1/ontologies_controller.rb
+++ b/app/controllers/api/v1/ontologies_controller.rb
@@ -1,6 +1,7 @@
 class Api::V1::OntologiesController < Api::V1::Base
   inherit_resources
   actions :show
+  before_filter :set_format
 
   respond_to :text, only: %i(show)
 
@@ -21,5 +22,9 @@ class Api::V1::OntologiesController < Api::V1::Base
   def version
     finder = OntologyVersionFinder.new(params[:reference], resource)
     @version ||= finder.find if params[:reference]
+  end
+
+  def set_format
+    request.format = :text unless request.format.to_sym == :json
   end
 end
--- a/app/controllers/api/v1/ontologies_controller.rb
+++ b/app/controllers/api/v1/ontologies_controller.rb
@@ -1,7 +1,7 @@
 class Api::V1::OntologiesController < Api::V1::Base
+  include PlainTextDefaultHelper
   inherit_resources
   actions :show
-  before_filter :set_format
 
   respond_to :text, only: %i(show)
 
@@ -22,9 +22,5 @@ class Api::V1::OntologiesController < Api::V1::Base
   def version
     finder = OntologyVersionFinder.new(params[:reference], resource)
     @version ||= finder.find if params[:reference]
-  end
-
-  def set_format
-    request.format = :text unless request.format.to_sym == :json
   end
 end
--- a/app/controllers/ontologies_controller.rb
+++ b/app/controllers/ontologies_controller.rb
@@ -52,7 +52,6 @@ class OntologiesController < InheritedResources::Base
 
   def show
     @content_object = :ontology
-    params[:id] = params[:ontology_id] if params[:ontology_id]
 
     if !params[:repository_id]
       # redirect for legacy routing
--- a/app/controllers/ontologies_controller.rb
+++ b/app/controllers/ontologies_controller.rb
@@ -52,6 +52,8 @@ class OntologiesController < InheritedResources::Base
 
   def show
     @content_object = :ontology
+    ref_id = params[:id]
+    params[:id] = params[:ontology_id] if params[:ontology_id]
 
     if !params[:repository_id]
       # redirect for legacy routing
@@ -120,6 +122,7 @@ class OntologiesController < InheritedResources::Base
       clazz = type=='DistributedOntology' ? DistributedOntology : SingleOntology
       @ontology = clazz.new params[:ontology]
       @ontology.repository = parent
+      puts @ontology
       @ontology
     end
   end
--- a/app/controllers/ontologies_controller.rb
+++ b/app/controllers/ontologies_controller.rb
@@ -52,7 +52,6 @@ class OntologiesController < InheritedResources::Base
 
   def show
     @content_object = :ontology
-    ref_id = params[:id]
     params[:id] = params[:ontology_id] if params[:ontology_id]
 
     if !params[:repository_id]
--- a/app/controllers/ontologies_controller.rb
+++ b/app/controllers/ontologies_controller.rb
@@ -121,7 +121,6 @@ class OntologiesController < InheritedResources::Base
       clazz = type=='DistributedOntology' ? DistributedOntology : SingleOntology
       @ontology = clazz.new params[:ontology]
       @ontology.repository = parent
-      puts @ontology
       @ontology
     end
   end
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -17,7 +17,7 @@ Specroutes.define(Ontohub::Application.routes) do
     end
 
   # Special (/ref-based) Loc/Id routes
-  specified_get '/ref/:reference/:repository_id/*locid' => 'ontologies#show',
+  specified_get '/ref/:reference/:repository_id/*locid' => 'ontology_versions#show',
     as: :ontology_iri_versioned,
     constraints: [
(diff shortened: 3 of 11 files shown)


ebolloff:
There are two types of errors in `Feature: Resolving LocId IRIs` and `Feature: LocId Routing`:
```
  Step: And I should have an API-command matching 'symbols' # features/step_definitions/loc_id_steps.rb:18
      expected "/repository-22/7388c85042a8c4f89fc2" to include "///symbols" (RSpec::Expectations::ExpectationNotMetError)
      ./features/step_definitions/loc_id_steps.rb:19:in `/^I should have an API\-command matching '([^']+)'$/'
      features/LocId.feature:7:in `And I should have an API-command matching 'symbols''
```
and
```
  Step: And the page title should include the subject's name # features/step_definitions/loc_id_routing_steps.rb:38
      undefined method `include?' for nil:NilClass (NoMethodError)
      ./features/step_definitions/loc_id_routing_steps.rb:39:in `/^the page title should include the subject's (\S+)$/'
      features/LocIdRouting.feature:7:in `And the page title should include the subject's name'
```

phyrog:
I believe this is because the test does not send the `text/html` accept header. I'll try to fix it.

phyrog:
Tests should be working now.

app/controllers/ontologies_controller.rb
```diff
@@ -52,6 +52,7 @@ def create
 
   def show
     @content_object = :ontology
+    params[:id] = params[:ontology_id] if params[:ontology_id]
```
phyrog: `params[:id]` is somehow set to `params[:reference]` instead of the `ontology_id`. The reference id can still be obtained by using `params[:reference]`

config/routes.rb
```diff
@@ -33,12 +31,14 @@
       BODY
     end
 
-  specified_get '/ref/:reference/:repository_id/*locid' => 'ontologies#show',
+  specified_get '/ref/:reference/:repository_id/*locid' => 'api/v1/ontology_versions#show',
```
phyrog: Not my fault, I just moved it.

phyrog: Not my fault. I just moved it.

app/controllers/api/v1/ontologies_controller.rb
```diff
@@ -22,4 +23,8 @@ def version
     finder = OntologyVersionFinder.new(params[:reference], resource)
     @version ||= finder.find if params[:reference]
   end
+
+  def set_format
```
phyrog: Should be moved to a Controller concern when porting to Rails 4. If there is an elegant solution for sharing this with other Controllers, that works with Rails 3, I'm all ears.

features/step_definitions/loc_id_routing_steps.rb
```diff
@@ -24,10 +24,12 @@
 end
 
 Given(/^I visit my subject's locid$/) do
+  Capybara.current_session.driver.header('Accept', 'text/html')
```
phyrog: Could be moved to a more Central place, so it applies to all requests. Unfortunately I don't know why rack-test is even used here or where it is initialized, since our env.rb sets the default driver to poltergeist.

config/routes.rb
0robustus1: You could try moving routes.rb to the excluded files-list. Since the file is written in a DSL i would argue that it doesn't need to adhere to the same standards as *normal* code.

app/controllers/ontologies_controller.rb
0robustus1: Mhm, when using the `RefLocIdRouterConstraint`, `params[:id]` can be set to the version_id of the ontology versions as defined per https://github.com/ontohub/ontohub/blob/staging/lib/router_constraints.rb#L128. I'd assume that this can happen in this route: https://github.com/ontohub/ontohub/blob/staging/config/routes.rb#L36-L48. That sadly doesn't reveal an elegant solve, but might explain things ;)

app/controllers/api/v1/ontologies_controller.rb
ebolloff: I guess it's okay. We'll change it when upgrading to Rails 4.

eugenk: I agree with @ebolloff, but please add a comment in the code like in app/fake_records/proof.rb:109

config/routes.rb
ebolloff: @phyrog could you add the `routes.rb` to the excluded files-list of Hound?

app/controllers/ontologies_controller.rb
eugenk: which URL did you call that only sets `params[:ontology_id]` and not `params[:id]`

eugenk: When I visit `http://ontohub.dev/default/partial_order`
My `params` hash is
```
{"controller"=>"ontologies", "action"=>"show", "repository_id"=>"default", "locid"=>"partial_order", "id"=>3}
```

(There is no `ontology_id`)

app/controllers/api/v1/ontologies_controller.rb
eugenk: Oh, does a simple Ruby module not work here?

config/routes.rb
```diff
@@ -17,14 +17,12 @@
     end
 
   # Special (/ref-based) Loc/Id routes
-  specified_get '/ref/:reference/:repository_id/*locid' => 'api/v1/ontology_versions#show',
+  specified_get '/ref/:reference/:repository_id/*locid' => 'ontologies#show',
```
eugenk: why `ontologies#show` and not `ontology_versions#show`?

features/step_definitions/loc_id_routing_steps.rb
ebolloff: Maybe the default driver is called at some point. Because we didn't set it to Poltergeist, maybe Rack is called? You could try adding `Capybara.default_driver = :poltergeist` in `env.rb`.

app/controllers/ontologies_controller.rb
phyrog: `'/ref/:reference/:repository_id/*locid' => 'ontologies#show'` this one. But since this route is not needed, it can be removed.

config/routes.rb
```diff
@@ -17,25 +17,12 @@
     end
 
   # Special (/ref-based) Loc/Id routes
-  specified_get '/ref/:reference/:repository_id/*locid' => 'ontologies#show',
```
eugenk: Please leave this route in this file, but change the first line to
```ruby
specified_get '/ref/:reference/:repository_id/*locid' => 'ontologies#show',
```

There is a slight difference to the behaviour after this commit:

When requesting text/html of http://ontohub.dev/default/Cat on this commit, the response is a download of the file *without* the file extension.

When requesting text/html of http://ontohub.dev/default/Cat on this commit, the response is a download of the file *with* the file extension.

config/routes.rb
```diff
@@ -17,25 +17,12 @@
...
-ontology version referenced by the {reference}.
-      BODY
-    end
-
   specified_get '/ref/:reference/:repository_id/*locid' => 'api/v1/ontology_versions#show',
     as: :ontology_iri_versioned,
     constraints: [
       RefLocIdRouterConstraint.new(Ontology, ontology: :ontology_id),
     ] do
+      accept 'text/html'
```
eugenk: Please remove this line again (and leave the above route in there)

eugenk:
:+1:

ebolloff:
:+1:

PR #1769, opened 2016-08-10
1476 re-anaylse old ontologies

### LINKED ISSUE #1476: re-anaylse old ontologies
On Ontohub, there are many ontologies that do not have loc/ids and/or do not show a class tree though they are in OWL. Re-analyse all Ontohub ontologies, starting with that ontology that has been analysed the longest time ago. This should be done in the background.

ebolloff:
This adds a rake task for analyzing ontologies and fixes #1476. The jobs for reanalyzing are in the queue `hets-migration` and not default `hets`, so the queue is not overloaded. There's some documentation for pushing jobs in another queue from @eugenk.


lib/tasks/ontology.rake
```diff
@@ -0,0 +1,24 @@
+namespace :ontology do
+  desc 'Reanalyze old ontologies'
+  task :reanalyze_all => :environment do
+    Ontology.where(parent_id: nil, present: true).find_each do |ontology|
```
eugenk: Please order the ontologies by updated-date:
```ruby
Ontology.where(parent_id: nil, present: true).order('updated_at desc').find_each do |ontology|
```

eugenk:
:+1: if you adjust the ordering and obey Hound.

eugenk:
:+1:

PR #1775, opened 2016-08-12
Handle error for ontologies that don't have a file any more.

eugenk:
This shall fix the rake task `ontology:reanalyze_all`. Sometimes files have been deleted and the ontology is still there. Then, the ontology must be "removed".


ebolloff:
:+1:

PR #1780, opened 2016-08-12
1728 Eager load everything in lib

### LINKED ISSUE #1728: Not everything gets loaded in rails c
Logged into the `rails c production` on ontohub.org, when I execute
```ruby
ActiveRecord::Base.logger.level = 1; threads = Repository.find_by_path('eugens-test-1').ontologies.where(parent_id: nil).map(&:current_version).map { |ov| Thread.new { ov.parse_full } }.each(&:join)
```
I get errors that indicate incomplete application loading. It randomly doesn't find classes in the top namespace, like `Semaphore`, `AccessTokenDeletionWorker` or `TimeoutWorker` (raising `NameError`s with `NameError: uninitialized constant ...`). As far as I remember, they all reside in the `lib` directory. I'm not 100% sure of it, though.

phyrog:
Should fix #1728. `lib` will now be eager loaded. For this to work, we must not load `environment_light` when `Rails` is already defined.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -18,7 +18,7 @@ module Ontohub
     # -- all .rb files in that directory are automatically loaded.
 
     # Custom directories with classes and modules you want to be autoloadable.
-    config.autoload_paths += %W(#{config.root}/lib)
+    config.eager_load_paths += %W(#{config.root}/lib)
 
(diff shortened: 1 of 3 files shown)


ebolloff:
:+1:

PR #1787, opened 2016-08-13
Fix requiring libs

eugenk:
Because of the eager loading that was introduced in #1780, some classes were loaded multiple times because we have `require` statements in our code. This shall fix it.

Also, Hets could not be started via the rake task that invoker uses because the light environment could not be loaded.


phyrog:
:+1:

PR #1796, opened 2016-08-14
Fix Sidekiq loading

eugenk:
When Sidekiq started, it had 100% CPU load and did not finish loading
the rb files. This is because of a bug is Sidetiq:
https://github.com/tobiassvn/sidetiq/issues/31#issuecomment-23780476

This commit is a workaround for the bug.


phyrog:
After the tests are green (which I expect them to be): :+1:

PR #1797, opened 2016-08-14
Set group write permissions on repository create

eugenk:
This shall fix the permission denied error for new repositories: Users should now be able to edit files via both git-push and web.


ebolloff:
:+1:

PR #1798, opened 2016-08-14
1668 remove locid for prover output

0robustus1:
This introduces the consistent use of `///prover_output` command-style loc-ids for the prover-output. 

It includes some adjustments to the developer-setup guide for Mac OS X.

It also introduces small error raising for `url_for` in the locid-case.

Hopefully it works. ;)


lib/extend/action_dispatch/routing/polymorphic_routes.rb
```diff
@@ -41,6 +41,7 @@ def ontohub_url_authority
       end
 
       def build_locid_url(base_url, record, commands, query_components)
+        raise ArgumentError.new("locid not set for #{record}") if !record.respond_to?(:locid) || record.locid.nil?
```
0robustus1: Should i care? truly asking...

PR #1806, opened 2017-02-23
Add repository premium flag and lots of fixes

### LINKED ISSUE #1499: flag for selected repositories
We have too much junk in Ontohub. There should be a flag for selected repositories. Unselected repos will only be displayed on explicit request, also when searching ontologies. There should be a possibility to suggest repos for selection.

eugenk:
This fixes lots of small problems I stumbled across when I uploaded some examples of the TPTP library. The commit messages should say everything about the problems.

This shall also fix #1499.

Note that travis is red because it can't find hets in the ppa.


eugenk:
Travis can't run cucumber. It's the same for me, locally:
```
cannot load such file -- sinatra/base
```

